Issue #16, opened 2024-11-05
ENH: add `sinc`


lucascolley:
x-ref https://github.com/scipy/scipy/pull/21783#discussion_r1829177161

asmeurer:
Does everyone use the normalized definition of sinc?

lucascolley:
The array libraries we test with in SciPy all follow NumPy in using the normalized definition.

We could add a `normalized=True` kwarg if the unnormalized use-case ever arose?

asmeurer:
Not unless that is already common. It doesn't look like NumPy has it. You can always divide by pi if you want an unnormalized sinc. 

My main concern is whether we would be deciding on a convention here, or whether that convention has effectively already been decided by the community.

lucascolley:
> My main concern is whether we would be deciding on a convention here, or whether that convention has effectively already been decided by the community.

My two cents is that this library is not trying to weigh in on conventions at all, it is just taking what is used out in the wild and making that accessible in the array-agnostic context. If multiple conventions are used for some function name we should support all common use cases. If/when an API is considered for the standard is when there is a decision on convention.

It sounds like @rgommers would roughly agree from https://github.com/data-apis/array-api-extra/pull/10#issuecomment-2439874054.

PR #20, opened 2024-11-14
ENH: add `sinc`

lucascolley:
Closes gh-16

### DIFF
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -429,5 +429,7 @@ def sinc(x: Array, /, *, xp: ModuleType) -> Array:
         err_msg = "`x` must have a real floating data type."
         raise ValueError(err_msg)
     # no scalars in `where` - array-api#807
-    y = xp.pi * xp.where(x, x, xp.asarray(xp.finfo(x.dtype).smallest_normal, dtype=x.dtype))
+    y = xp.pi * xp.where(
+        x, x, xp.asarray(xp.finfo(x.dtype).smallest_normal, dtype=x.dtype)
+    )
     return xp.sin(y) / y
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -370,6 +370,8 @@ def sinc(x: Array, /, *, xp: ModuleType) -> Array:
     x : array
         Array (possibly multi-dimensional) of values for which to calculate
         ``sinc(x)``. Must have a real floating point dtype.
+    xp : array_namespace
+        The standard-compatible namespace for `x`.
 
     Returns
     -------
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -367,9 +367,9 @@ def sinc(x: Array, /, *, xp: ModuleType) -> Array:
 
     Parameters
     ----------
-    x : array
+    x : array of floats
         Array (possibly multi-dimensional) of values for which to calculate
-        ``sinc(x)``.
+        ``sinc(x)``. Should have a floating point dtype.
 
     Returns
     -------
@@ -423,5 +423,8 @@ def sinc(x: Array, /, *, xp: ModuleType) -> Array:
            -3.89817183e-17], dtype=array_api_strict.float64)
 
     """
+    if not xp.isdtype(x.dtype, "real floating"):
+        err_msg = "`x` must have a real floating data type."
+        raise ValueError(err_msg)
     y = xp.pi * xp.where(x == 0, xp.asarray(1.0e-20), x)
     return xp.sin(y) / y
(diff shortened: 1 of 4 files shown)


lucascolley:
cc @ev-br FYI

src/array_api_extra/_funcs.py
```diff
@@ -348,3 +348,83 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType) -> Array:
...
+            6.68206631e-02,  3.89817183e-17,
+           -5.84680802e-02, -8.90384387e-02,
+           -8.40918587e-02, -4.92362781e-02,
+           -3.89817183e-17], dtype=array_api_strict.float64)
+
+    """
+    if not xp.isdtype(x.dtype, "real floating"):
+        err_msg = "`x` must have a real floating data type."
+        raise ValueError(err_msg)
+    y = xp.pi * xp.where(x == 0, xp.asarray(1.0e-20), x)
```
jakirkham, line 429: Couple of thoughts

Would it make sense to rewrite this so the comparison is unneeded?

Also would it make sense to just `finfo` to grab `smallest_normal` or some other appropriately small number?

Considering both of these, we might end up with something like

```suggestion
    y = xp.pi * xp.where(x, x, xp.finfo(x.dtype).smallest_normal)
```

lucascolley: sounds good to me! (just grabbed this from NumPy)

src/array_api_extra/_funcs.py
```diff
@@ -348,3 +348,83 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType) -> Array:
...
+
+    The sinc function is equal to :math:`\sin(\pi x)/(\pi x)` for any argument
+    :math:`x\ne 0`. ``sinc(0)`` takes the limit value 1, making ``sinc`` not
+    only everywhere continuous but also infinitely differentiable.
+
+    .. note::
+
+        Note the normalization factor of ``pi`` used in the definition.
+        This is the most commonly used definition in signal processing.
+        Use ``sinc(x / np.pi)`` to obtain the unnormalized sinc function
```
lucascolley, line 365:
```suggestion
        Use ``sinc(x / xp.pi)`` to obtain the unnormalized sinc function
```

src/array_api_extra/_funcs.py
```diff
@@ -348,3 +348,83 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType) -> Array:
...
+            6.68206631e-02,  3.89817183e-17,
+           -5.84680802e-02, -8.90384387e-02,
+           -8.40918587e-02, -4.92362781e-02,
+           -3.89817183e-17], dtype=array_api_strict.float64)
+
+    """
+    if not xp.isdtype(x.dtype, "real floating"):
+        err_msg = "`x` must have a real floating data type."
+        raise ValueError(err_msg)
+    y = xp.pi * xp.where(x, x, xp.finfo(x.dtype).smallest_normal)
```
lucascolley, line 429:
```suggestion
    # no scalars in `where` - array-api#807
    y = xp.pi * xp.where(x, x, xp.asarray(xp.finfo(x.dtype).smallest_normal))
```

src/array_api_extra/_funcs.py
jakirkham: Seems like a reasonable starting point

Perhaps we can follow up in NumPy after

src/array_api_extra/_funcs.py
jakirkham: Ah good point. Thanks for fixing that! 🙏

src/array_api_extra/_funcs.py
```diff
@@ -348,3 +348,86 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType) -> Array:
...
+           -5.84680802e-02, -8.90384387e-02,
+           -8.40918587e-02, -4.92362781e-02,
+           -3.89817183e-17], dtype=array_api_strict.float64)
+
+    """
+    if not xp.isdtype(x.dtype, "real floating"):
+        err_msg = "`x` must have a real floating data type."
+        raise ValueError(err_msg)
+    # no scalars in `where` - array-api#807
+    y = xp.pi * xp.where(x, x, xp.asarray(xp.finfo(x.dtype).smallest_normal))
```
jakirkham, line 432: Probably worth matching `x.dtype` when using this value to avoid accidental type promotion (and additional memory usage)

```suggestion
    y = xp.pi * xp.where(x, x, xp.asarray(xp.finfo(x.dtype).smallest_normal, dtype=x.dtype))
```

lucascolley:
the rendered docs look good to me, let's pull this in - thanks for the review @jakirkham !

jakirkham:
Thanks Lucas! 🙏

Happy to help 🙂

PR #37, opened 2024-11-26
TYP: adopt based{mypy, pyright}

lucascolley:
@jorenham could you review :) ?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ numpy = "*"
 pytest = "*"
 
 [tool.pixi.feature.lint.pypi-dependencies]
-basedmypy = "*"
+basedmypy = { version = "*", extras = ["faster-cache"] }
 basedpyright = "*"
 
 [tool.pixi.feature.lint.tasks]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,7 +169,7 @@ enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 warn_unreachable = true
 disallow_untyped_defs = false
 disallow_incomplete_defs = false
-# array-api#589
+# data-apis/array-api#589
 disallow_any_expr = false
 
 [[tool.mypy.overrides]]
@@ -186,7 +186,7 @@ pythonVersion = "3.10"
 pythonPlatform = "All"
 typeCheckingMode = "strict"
 
-# array-api#589
+# data-apis/array-api#589
 reportAny = false
 reportExplicitAny = false
 reportUnknownMemberType = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,19 +70,23 @@ array-api-extra = { path = ".", editable = true }
 
 [tool.pixi.feature.lint.dependencies]
 pre-commit = "*"
-mypy = "*"
 pylint = "*"
 # import dependencies for mypy:
 array-api-strict = "*"
 numpy = "*"
 pytest = "*"
 
+[tool.pixi.feature.lint.pypi-dependencies]
+basedmypy = "*"
+basedpyright = "*"
+
 [tool.pixi.feature.lint.tasks]
 pre-commit-install = { cmd = "pre-commit install" }
 pre-commit = { cmd = "pre-commit run -v --all-files --show-diff-on-failure" }
 mypy = { cmd = "mypy", cwd = "." }
 pylint = { cmd = ["pylint", "array_api_extra"], cwd = "src" }
-lint = { depends-on = ["pre-commit", "pylint", "mypy"] }
+pyright = { cmd = "basedpyright", cwd = "." }
+lint = { depends-on = ["pre-commit", "pylint", "mypy", "pyright"] }
 
 [tool.pixi.feature.tests.dependencies]
 pytest = ">=6"
@@ -165,11 +169,27 @@ enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 warn_unreachable = true
 disallow_untyped_defs = false
 disallow_incomplete_defs = false
+# array-api#589
+disallow_any_expr = false
 
 [[tool.mypy.overrides]]
 module = "array_api_extra.*"
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
+
+
+# pyright
+
+[tool.pyright]
+include = ["src", "tests"]
+pythonVersion = "3.10"
+pythonPlatform = "All"
+typeCheckingMode = "strict"
+
+# array-api#589
+reportAny = false
+reportExplicitAny = false
+reportUnknownMemberType = false
 
 
 # Ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,11 +180,11 @@ disallow_incomplete_defs = true
 
 # pyright
 
-[tool.pyright]
+[tool.basedpyright]
 include = ["src", "tests"]
 pythonVersion = "3.10"
 pythonPlatform = "All"
-typeCheckingMode = "strict"
+typeCheckingMode = "all"
 
 # data-apis/array-api#589
 reportAny = false
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -5,7 +5,7 @@ import typing
 import warnings
 
 # array-api-strict#6
-import array_api_strict as xp  # type: ignore[import-untyped]
+import array_api_strict as xp  # type: ignore[import-untyped]  # pyright: ignore[reportMissingTypeStubs]
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_array_equal, assert_equal
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -13,7 +13,8 @@ from numpy.testing import assert_allclose, assert_array_equal, assert_equal
 from array_api_extra import atleast_nd, cov, create_diagonal, expand_dims, kron, sinc
 
 if TYPE_CHECKING:
-    Array = Any  # To be changed to a Protocol later (see array-api#589)
+    # To be changed to a Protocol later (see array-api#589)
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -165,13 +169,29 @@ enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 warn_unreachable = true
 disallow_untyped_defs = false
 disallow_incomplete_defs = false
```
lucascolley, line 171: there may be spurious options in the `mypy` config now (given changed defaults)

jorenham:
Any thoughts on this @KotlinIsland ?

pyproject.toml
```diff
@@ -70,19 +70,23 @@ array-api-extra = { path = ".", editable = true }
...
+basedmypy = "*"
+basedpyright = "*"
+
 [tool.pixi.feature.lint.tasks]
 pre-commit-install = { cmd = "pre-commit install" }
 pre-commit = { cmd = "pre-commit run -v --all-files --show-diff-on-failure" }
 mypy = { cmd = "mypy", cwd = "." }
 pylint = { cmd = ["pylint", "array_api_extra"], cwd = "src" }
-lint = { depends-on = ["pre-commit", "pylint", "mypy"] }
+pyright = { cmd = "basedpyright", cwd = "." }
```
jorenham, line 88: this should probably be added to the CI workflow

pyproject.toml
```diff
@@ -70,19 +70,23 @@ array-api-extra = { path = ".", editable = true }
...
 pre-commit = "*"
-mypy = "*"
 pylint = "*"
 # import dependencies for mypy:
 array-api-strict = "*"
 numpy = "*"
 pytest = "*"
 
+[tool.pixi.feature.lint.pypi-dependencies]
+basedmypy = "*"
```
jorenham, line 80: you could use the `basedmypy[faster-cache]` extra to speed it up a bit

pyproject.toml
```diff
@@ -165,13 +169,29 @@ enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
...
 disallow_incomplete_defs = true
 
 
+# pyright
+
+[tool.pyright]
+include = ["src", "tests"]
+pythonVersion = "3.10"
+pythonPlatform = "All"
+typeCheckingMode = "strict"
```
jorenham, line 187: You could also try setting this to `"recommended"` or `"all"`, which unleases the full power of basedpyright.
See https://docs.basedpyright.com/latest/configuration/config-files/#diagnostic-settings-defaults for what this exactly enables

pyproject.toml
```diff
@@ -165,13 +169,29 @@ enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
...
+
+# data-apis/array-api#589
+reportAny = false
+reportExplicitAny = false
+reportUnknownMemberType = false
```
jorenham, line 192: Instead of disabling this globally, you could also consider using the `baseline` feature:
https://docs.basedpyright.com/latest/benefits-over-pyright/baseline/

tests/test_funcs.py
```diff
@@ -13,7 +13,8 @@
 from array_api_extra import atleast_nd, cov, create_diagonal, expand_dims, kron, sinc
 
 if TYPE_CHECKING:
-    Array = Any  # To be changed to a Protocol later (see array-api#589)
+    # To be changed to a Protocol later (see data-apis/array-api#589)
+    Array = Any  # type: ignore[no-any-explicit]
```
jorenham, line 17: also defined in `_typing`

tests/test_funcs.py
```diff
@@ -197,7 +198,7 @@ def test_kron_smoke(self):
             ((2, 0, 0, 2), (2, 0, 2)),
         ],
     )
-    def test_kron_shape(self, shape_a, shape_b):
+    def test_kron_shape(self, shape_a: tuple[int], shape_b: tuple[int]):
```
jorenham, line 201: a `tuple[int]` only accepts tuples of length 1

```suggestion
    def test_kron_shape(self, shape_a: tuple[int, ...], shape_b: tuple[int, ...]):
```

tests/test_funcs.py
```diff
@@ -271,7 +272,7 @@ def test_simple(self):
         assert_allclose(w, xp.flip(w, axis=0))
 
     @pytest.mark.parametrize("x", [0, 1 + 3j])
-    def test_dtype(self, x):
+    def test_dtype(self, x: int | complex):
```
jorenham, line 275: Perhaps you're already aware of this; but in the static typing word, this is equivalent to `complex` (even though it isn't at runtime...). But you could also argue that it's better to be explicit about it, and type it as `int | complex` instead of `complex` like you've currently done.

pyproject.toml
```diff
@@ -70,19 +70,23 @@ array-api-extra = { path = ".", editable = true }
...
+basedpyright = "*"
+
 [tool.pixi.feature.lint.tasks]
 pre-commit-install = { cmd = "pre-commit install" }
 pre-commit = { cmd = "pre-commit run -v --all-files --show-diff-on-failure" }
 mypy = { cmd = "mypy", cwd = "." }
 pylint = { cmd = ["pylint", "array_api_extra"], cwd = "src" }
-lint = { depends-on = ["pre-commit", "pylint", "mypy"] }
+pyright = { cmd = "basedpyright", cwd = "." }
+lint = { depends-on = ["pre-commit", "pylint", "mypy", "pyright"] }
```
jorenham, line 89: you could also consider adding mypy and pyright to the pre-commit config:

https://github.com/pre-commit/mirrors-mypy
https://docs.basedpyright.com/latest/installation/pre-commit%20hook/

pyproject.toml
lucascolley: > I was mostly talking about the reportUnknownMemberType

These are all about the types coming from array-api-strict I think:

`/array-api-extra/tests/test_funcs.py:183:13 - error: Type of "ones" is partially unknown
    Type of "ones" is "(shape: int | Tuple[int, ...], *, dtype: _DType | None = None, device: Unknown | None = None) -> Array" (reportUnknownMemberType)`

If so, I think more of those will continue to be introduced for the time being, so I'd rather ignore globally.

jorenham: But even so, it's probably best to baseline those errors, than to globally hide errors about unknown member types, IMO

jorenham: Alternatively, you could als try to create some quick 'n dirty stubs for it

lucascolley: I'll leave it for this PR and open an issue 👍

pyproject.toml
lucascolley: Switching to `all` introduces one error:

`array-api-extra/tests/test_funcs.py:8:8 - error: Stub file not found for "array_api_strict" (reportMissingTypeStubs)`

pyproject.toml
jorenham: I believe those are all documented here: https://kotlinisland.github.io/basedmypy/error_code_list.html

lucascolley:
thanks Joren!

tests/test_funcs.py
KotlinIsland: i HATE tuple semantics

tests/test_funcs.py
asmeurer: This little issue is going to be something you're going to have to decide how to deal with in typing the array API. 

In the array API (and the numerics world in general), integers and floats are *not* compatible types. The array API specifically discourages implicitly casting from one to the other. But according to [Python typing](https://peps.python.org/pep-0484/#the-numeric-tower), any function that accepts floats should also accept integers. It's going to be even more of an issue now that we are adding support for Python scalars to functional APIs (https://github.com/data-apis/array-api/issues/807). There's plenty of functions that accept float but are not required to accept int (and even some that *must not* accept int). See also https://github.com/data-apis/array-api/pull/862#discussion_r1857364664.

jorenham: I've ran into this issue more times than I can count...
Recently I found a workaround for (at least) `int` that works with both mypy and pyright, which is available in [`optype`](https://github.com/jorenham/optype) as `optype.typing.JustInt` ([docs](https://github.com/jorenham/optype/blob/master/README.md#just-types)).
There's also the more general `optype.typing.Just`, but because of an issue in typeshed (which recently has been fixed), and a bug in mypy, this currently doesn't work for `float` and `complex`. If the next pyright release bundles the latest typeshed, then it should at least be possible in pyright to use `optype.typing.Just[float]` so that it rejects `int` and `bool` and "just" accepts `float` instances.

KotlinIsland: i'm planning to address this situation entirely in the basedtyping ecosystem, perhaps not so relevant to the 484 ecosystem

pyproject.toml
KotlinIsland:
```toml
warn_unused_configs = true
strict = true
enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
warn_unreachable = true
```

these are all enabled by default in basedmypy

tests/test_funcs.py
rgommers: Is it really an issue? The `int |` was kept in a number of places for clarity, but we had the discussion between about `int | float` and `float` meaning the same in type annotation.

Note that it's _very_ different between array library dtypes and Python scalars. For Python scalars the considerations that made us avoid mixed int-float promotion rules all just don't exist.

jorenham: Yea it's a pretty big issue, especially for numpy, see https://github.com/numpy/numpy/issues/27032

jorenham: But apart from the issue of the overlapping overloads, the worst thing is that it's type-unsafe. For example

```py
def get_hex(x: float, /) -> str:
    return x.hex()

get_hex(42)  # accepted
```

which at runtime results in

```
AttributeError: 'int' object has no attribute 'hex'
```

PR #47, opened 2024-12-06
MAINT: depend on array-api-compat

### LINKED ISSUE #6: ENH/API: xp-bound namespaces, array-api-compat
Currently, functions of this package require passing a standard-compatible namespace as `xp=xp`. This works fine, but there have been suggestions that it might be nice to avoid this requirement. There are at least a few ways we could go about this:

### (1) `xpx.bind_namespace`

Usage:

```python
import array_api_strict as xpx
...
xp = array_namespace(x)
xpx = xpx.bind_namespace(xp)
x = xpx.atleast_nd(x, ndim=2)
y = xp.sum(x)
z = xpx.some_func(y)
```

A potential implementation:

```python
extra_funcs = {'atleast_nd': atleast_nd, ...}

def bind_namespace(xp: ModuleType) -> ModuleType:
    class BoundNamespace:
        def __getattr__(self, name: str):
            if name in extra_funcs:
                return functools.partial(extra_funcs[name], xp=xp)
            else:
               return AttributeError(...)

    return BoundNamespace(xp)
```

I like this idea. If we encounter use cases where a library wants to use multiple `xpx` functions in the same local scope and finds the `xp=xp` pattern too cumbersome, I think we should add this. I think we can leave it out for now until that situation arises.

### (2) `xpx.extra_namespace`

Usage:

```python
import array_api_strict as xpx
...
xp = array_namespace(x)
xpx = xpx.extra_namespace(xp)
x = xpx.atleast_nd(x, ndim=2)
y = xpx.sum(x)  # XXX: xpx instead of xp
z = xpx.some_func(y)
```

A potential implementation:

```python
extra_funcs = {'atleast_nd': atleast_nd, ...}

def extra_namespace(xp: ModuleType) -> ModuleType:
    class ExtraNamespace:
        def __getattr__(self, name: str):
            if name in extra_funcs:
                return functools.partial(extra_funcs[name], xp=xp)
            else:
               return getattr(xp, name)  # XXX: delegate to xp instead of error

    return ExtraNamespace(xp)
```

I would not want to add this yet. I think we should keep separation between the standard namespace and the 'extra' namespace, at least until this library matures.

### (3) Use `array_api_compat.array_namespace` internally

This would provide the most flexible API and be the least LOC to use. One could use `xpx` functions on standard-incompatible arrays, and let array-api-compat handle the compatibility, without having to pass an `xp` argument.

We don't yet have a use case where it is clearly beneficial to be able to pass standard-incompatible arrays. Consumer libraries using array-api-extra would already be computing with standard-compatible arrays internally. I don't see the need to support the following use case:

```python
import torch
import array_api_strict as xpx
...
x = torch.asarray([1, 2, 3])
xpx.some_func(x)             # works
torch.some_standard_func(x)  # does not work
```

Another complication is that consumer libraries like SciPy wrap `array_namespace` to provide custom behaviour for scalars and other types. We would want the internal `array_namespace` to be the consumer library's wrapped version rather than the base one from array-api-compat.

I'm also not sure that the 1 LOC save over option (1) of this post for standard-compatible arrays is worth introducing a dependency on array-api-compat.

Overall, this would complicate things a lot with situations of co-vendoring array-api-compat and array-api-extra, which is the primary use-case for the library right now. This might be a better idea in the future if a need for handling standard-incompatible arrays arises (for example, if one wants to use functions from `xpx` with just a single library).

crusaderky:
- Closes #6
- Blocked by #48
- Blocks implementation of `at` (https://github.com/data-apis/array-api-compat/pull/205)

Follow-up: vendoring hook in scipy

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ This project exists thanks to the following contributors
     </tr>
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://ogrisel.com"><img src="https://avatars.githubusercontent.com/u/89061?v=4?s=100" width="100px;" alt="Olivier Grisel"/><br /><sub><b>Olivier Grisel</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=ogrisel" title="Code">💻</a></td>
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="Guido Imperiale"/><br /><sub><b>Guido Imperiale</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="crusaderky"/><br /><sub><b>crusaderky</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
     </tr>
   </tbody>
 </table>
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ This project exists thanks to the following contributors
     </tr>
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://ogrisel.com"><img src="https://avatars.githubusercontent.com/u/89061?v=4?s=100" width="100px;" alt="Olivier Grisel"/><br /><sub><b>Olivier Grisel</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=ogrisel" title="Code">💻</a></td>
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="crusaderky"/><br /><sub><b>crusaderky</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="Guido Imperiale"/><br /><sub><b>Guido Imperiale</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
     </tr>
   </tbody>
 </table>
--- a/docs/index.md
+++ b/docs/index.md
@@ -51,16 +51,57 @@ If you require stability, it is recommended to pin `array-api-extra` to
 a specific version, or vendor the library inside your own.
 ```
 
+```{note}
+This library depends on array-api-compat. We aim for compatibility with
+the latest released version of array-api-compat, and your mileage may vary
+with older or dev versions.
+```
+
 (vendoring)=
 
 ## Vendoring
 
 To vendor the library, clone
-[the repository](https://github.com/data-apis/array-api-extra) and copy it into
-the appropriate place in your library, like:
+[the array-api-extra repository](https://github.com/data-apis/array-api-extra)
+and copy it into the appropriate place in your library, like:
 
 ```
-cp -R array-api-extra/ mylib/vendored/array_api_extra
+cp -a array-api-extra/src/array_api_extra mylib/vendored/
+```
+
+`array-api-extra` depends on `array-api-compat`. You may either add a dependency
+in your own project to `array-api-compat` or vendor it too:
+
+1. Clone
+   [the array-api-compat repository](https://github.com/data-apis/array-api-compat)
+   and copy it next to your vendored array-api-extra:
+
+   ```
+   cp -a array-api-compat/array_api_compat mylib/vendored/
+   ```
+
+2. Create a new hook file which array-api-extra will use instead of the
+   top-level `array-api-compat` if present:
+
+   ```
+   echo 'from mylib.vendored.array_api_compat import *' > mylib/vendored/_array_api_compat_vendor.py
+   ```
+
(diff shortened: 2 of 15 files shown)


lucascolley:
> WIP: vendoring hook in scipy (if you merge this now, scipy will break!)
> 
> Note that this will break all third-party libraries that vendor array-api-extra through a unpinned git subproject, like scipy does.

Why? I was under the impression that "unpinned" git submodules are not bumped unless someone deliberately does so. All of SciPy's git submodules are unpinned.

crusaderky:
> I was under the impression that "unpinned" git submodules are not bumped unless someone deliberately does so

I wasn't aware of that, good to know

lucascolley:
I think we will want to document somewhere what we are relying on from array-api-compat, or otherwise indicate which versions of array-api-compat this can depend on or be co-vendored with.

I think I'm fine with something along the lines of "we aim for compatibility with the latest released version of array-api-compat, and your mileage may vary with older or dev version".

crusaderky:
@lucascolley this is ready for review. Could you please enable workflows for this PR?

tests/test_funcs.py
```diff
@@ -28,91 +28,96 @@ class TestAtLeastND:
...
     def test_device(self):
         device = xp.Device("device1")
         x = xp.asarray([1, 2, 3], device=device)
-        assert atleast_nd(x, ndim=2, xp=xp).device == device
+        assert atleast_nd(x, ndim=2).device == device
+
+    def test_xp(self):
+        x = xp.asarray(1)
+        y = atleast_nd(x, ndim=0, xp=xp)
+        assert_array_equal(y, x)
```
crusaderky, line 96: (here and in all other `test_xp`) I couldn't come up with an interesting use case where an explicit xp yields different output from xp=None?

src/array_api_extra/_funcs.py
```diff
@@ -28,8 +29,8 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType) -> Array:
     x : array
     ndim : int
         The minimum number of dimensions for the result.
-    xp : array_namespace
-        The standard-compatible namespace for `x`.
+    xp : array_namespace, optional
+        The standard-compatible namespace for `x`. Default: infer
```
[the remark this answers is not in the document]
lucascolley: Yes, it avoids redundant calls to `array_namespace`. In SciPy we are likely to want to keep passing `xp=xp` where we already have `xp`.

tests/test_funcs.py
lucascolley: Yep, it shouldn't affect output, since this library assumes inputs are already arrays.

crusaderky:
Could you please unlock this?
image

crusaderky:
I executed `rm pixi.lock && pixi run tests` to update the lock; not sure if it's the correct process?

lucascolley:
> rm pixi.lock && pixi run tests 

`pixi update` should do

> Could you please unlock this?

Unfortunately, I believe manual approval has to be given every time, for first-time contributors. Feel free to submit a separate PR with a trivial change if you want to unblock that.

lucascolley:
looks like you may want to bump the version of `pixi` used by the CI workflows, if your local version is newer

crusaderky:
> Feel free to submit a separate PR with a trivial change if you want to unblock that.

https://github.com/data-apis/array-api-extra/pull/48

crusaderky:
I've rebased this PR on top of #48 to get a more readable pixi.lock diff

src/array_api_extra/_lib/_compat.pyi
```diff
@@ -0,0 +1,13 @@
...
+
+class ArrayModule(ModuleType):
+    def device(self, x: Array, /) -> Device: ...
+
+def array_namespace(
+    *xs: Array,
+    api_version: str | None = None,
+    use_compat: bool | None = None,
+) -> ArrayModule: ...
+def device(x: Array, /) -> Device: ...
```
crusaderky, line 13: This is a temporary hack to make mypy AND pyright happy. In the long run we should make array_api_compat typed.

crusaderky:
@lucascolley all green; ready for final review and merge

vendor_tests/_array_api_compat_vendor.py
```diff
@@ -0,0 +1,9 @@
+# This file is a hook imported by src/array_api_extra/_lib/_compat.py
+from __future__ import annotations
+
+from .array_api_compat import *  # noqa: F403
+from .array_api_compat import array_namespace as array_namespace_compat
```
lucascolley, line 5: why do we want to expose this as `array_namespace_compat` too?

docs/index.md
```diff
@@ -51,16 +51,57 @@ If you require stability, it is recommended to pin `array-api-extra` to
...
+from mylib.vendored.array_api_compat import array_namespace as _array_namespace_orig
+
+
+def array_namespace(*xs, **kwargs):
+    import mylib
+
+    if any(isinstance(x, mylib.MyArray) for x in xs):
+        return mylib
+    else:
+        return _array_namespace_orig(*xs, **kwargs)
```
lucascolley, line 104: the changes in this file look great, thanks!

lucascolley (approved):
thanks a lot @crusaderky, this looks great!

I suppose the scope of this package expanding is a blessing and a curse :). We will need to update the 'scope' section of the docs, but I'm happy to tackle this separately.

vendor_tests/_array_api_compat_vendor.py
crusaderky: Because we're overriding it immediately below. We're overriding it so that we can safely test `is` in the unit test. Adding a comment.

vendor_tests/__init__.py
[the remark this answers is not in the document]
crusaderky: Yes, it allows for relative imports, which in turn is necessary to distinguish the (potentially) globally-installed array-api-extra and array-api-compat from the vendored ones

crusaderky:
@lucascolley all comments addressed; this should be ready for merge

README.md
```diff
@@ -66,7 +66,7 @@ This project exists thanks to the following contributors
...
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="crusaderky"/><br /><sub><b>crusaderky</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/crusaderky"><img src="https://avatars.githubusercontent.com/u/6213168?v=4?s=100" width="100px;" alt="Guido Imperiale"/><br /><sub><b>Guido Imperiale</b></sub></a><br /><a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Code">💻</a> <a href="https://github.com/data-apis/array-api-extra/commits?author=crusaderky" title="Documentation">📖</a></td>
```
lucascolley, line 69: You may also want to update `.allcontributors-rc` and `CONTRIBUTORS.md`

crusaderky: Moved to separate PR https://github.com/data-apis/array-api-extra/pull/52

lucascolley:
thanks!

Issue #17, opened 2024-11-07
ENH: add testing support utilities


ev-br:
SciPy has two sets of testing support utilities:

- `{skip_xfail}_xp_backends` skips, https://github.com/scipy/scipy/blob/main/scipy/conftest.py#L215 . These are convenient for testing across the range of array api backends. Note that these are pytest-specific.
- assertions, `xp_assert_{equal,close}`, https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L299. These wrap backend-specific assertions, and scipy effectively transitions from `np.testing` to these ones. Quite a bit of effort went into crafting these ones, and ISTM these can be useful across the ecosystem.

lucascolley:
These utilities are specific to the libraries we test with in SciPy, so I don't think they would fit into the current scope of this library.

I think it may be possible to generalise `{skip_xfail}_xp_backends` over devices and skip based on https://data-apis.org/array-api/draft/API_specification/generated/array_api.info.capabilities.html rather than individual backends. But we would probably still want something separate in SciPy which is able to deal with specific implementations.

Since there are no `xp.testing` primitives in the standard, I don't see how we would add `xp_assert_{equal,close}` in an agnostic way.

I agree that these utilities should be pulled out of SciPy into some package which other consuming libraries can re-use. But I think that it would be useful to keep them separate from array-api-extra to avoid confusion, by keeping this library as purely building on top of the standard.

asmeurer:
I discussed this with Evgeni a few weeks ago, and based on his description of the SciPy tooling, I suggested that these utilities be split out into a separate package that is a pytest extension (e.g., called something like `pytest-arrays` or `pytest-array-api`).

ev-br:
To gauge interest of other downstream adopters: @betatim would you consider using scipy's `@skip_xp_backends` machinery and/or backend-agnostic assertions, `xp_assert_{close,equal}` in your scikit-learn work if they were in a separate package, as suggested above?
(We can take this discussion elsewhere if that'd be more convenient)

betatim:
A bit tricky to judge in the abstract.

Right now we convert all outputs back to Numpy arrays and then use the "standard" all close, equal, etc tooling. So in principle having a `xp_assert_close` could be nice, but in practice we have already solved this problem. And the current implementation isn't causing us any trouble - don't change what ain't broken?

[...]

PR #53, opened 2024-12-09
ENH: implement `at`

crusaderky:
Implement a new `at(x, idx)` or `at(x)[idx]` function, mocking the syntax of [JAX's omonymous method](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html) .
This is propaedeutic to JAX support in libraries that support the Array API, e.g. scipy.

Moved from https://github.com/data-apis/array-api-compat/pull/205

# Blockers
- https://github.com/data-apis/array-api-compat/pull/205
- https://github.com/data-apis/array-api-compat/pull/207
- https://github.com/data-apis/array-api-compat/pull/211
# Blocks
- https://github.com/scipy/scipy/pull/22070

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,8 +62,8 @@ channels = ["https://prefix.dev/conda-forge"]
 platforms = ["linux-64", "osx-arm64", "win-64"]
 
 [tool.pixi.dependencies]
-python = ">=3.10.15,<3.14"
-array-api-compat = ">=1.10.0"
+python = ">=3.10.16,<3.14"
+array-api-compat = ">=1.10.0,<2"
 
 [tool.pixi.pypi-dependencies]
 array-api-extra = { path = ".", editable = true }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dynamic = ["version"]
-# dependencies = ["array-api-compat>=1.10.0"]  # Do not release
+dependencies = ["array-api-compat>=1.10.0"]
 
 [project.optional-dependencies]
 tests = [
@@ -63,11 +63,9 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
 
 [tool.pixi.dependencies]
 python = ">=3.10.15,<3.14"
-# array-api-compat = ">=1.10.0"  # Do not release
+array-api-compat = ">=1.10.0"
 
 [tool.pixi.pypi-dependencies]
-# Do not release: main at least @ gh#205
-array-api-compat = { git = "https://github.com/data-apis/array-api-compat.git" }
 array-api-extra = { path = ".", editable = true }
 
 [tool.pixi.feature.lint.dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dynamic = ["version"]
-dependencies = ["array-api-compat>=1.1.1"]
+# dependencies = ["array-api-compat>=1.10.0"]  # Do not release
 
 [project.optional-dependencies]
 tests = [
@@ -63,9 +63,11 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
 
 [tool.pixi.dependencies]
 python = ">=3.10.15,<3.14"
-array-api-compat = ">=1.1.1"
+# array-api-compat = ">=1.10.0"  # Do not release
 
 [tool.pixi.pypi-dependencies]
+# Do not release: main at least @ gh#205
+array-api-compat = { git = "https://github.com/data-apis/array-api-compat.git" }
 array-api-extra = { path = ".", editable = true }
 
 [tool.pixi.feature.lint.dependencies]
@@ -130,6 +132,35 @@ python = "~=3.10.0"
 [tool.pixi.feature.py313.dependencies]
 python = "~=3.13.0"
 
+# Backends that can run on CPU-only hosts
+[tool.pixi.feature.backends.target.linux-64.dependencies]
+pytorch = "*"
+dask = "*"
+sparse = ">=0.15"
+jax = "*"
+
+[tool.pixi.feature.backends.target.osx-arm64.dependencies]
+pytorch = "*"
+dask = "*"
+sparse = ">=0.15"
+jax = "*"
+
+[tool.pixi.feature.backends.target.win-64.dependencies]
+# pytorch = "*"  # Package unavailable on Windows
+dask = "*"
+sparse = ">=0.15"
+# jax = "*"  # Package unavailable on Windows
+
+# Backends that require a GPU host and a CUDA driver
+[tool.pixi.feature.cuda-backends.target.linux-64.dependencies]
+cupy = "*"
+
+[tool.pixi.feature.cuda-backends.target.osx-arm64.dependencies]
+# cupy = "*"  # Package unavailable on macOSX
+
+[tool.pixi.feature.cuda-backends.target.win-64.dependencies]
+cupy = "*"
+
 [tool.pixi.environments]
 default = { solve-group = "default" }
 lint = { features = ["lint"], solve-group = "default" }
@@ -138,7 +169,9 @@ docs = { features = ["docs"], solve-group = "default" }
 dev = { features = ["lint", "tests", "docs", "dev"], solve-group = "default" }
 ci-py310 = ["py310", "tests"]
 ci-py313 = ["py313", "tests"]
-
+# CUDA not available on free github actions
+ci-backends = ["py310", "tests", "backends"]
+tests-backends = ["py310", "tests", "backends", "cuda-backends"]
 
 # pytest
 
@@ -195,6 +228,8 @@ reportAny = false
 reportExplicitAny = false
 # data-apis/array-api-strict#6
 reportUnknownMemberType = false
+# no array-api-compat type stubs
+reportUnknownVariableType = false
 
 
 # Ruff
@@ -236,6 +271,7 @@ ignore = [
     "PLR09",   # Too many <...>
     "PLR2004", # Magic value used in comparison
     "ISC001",  # Conflicts with formatter
+    "N801",    # Class name should use CapWords convention
     "N802",    # Function name should be lowercase
     "N806",    # Variable in function should be lowercase
 ]
@@ -271,6 +307,7 @@ checks = [
     "ES01",
 ]
 exclude = [ # don't report on objects that match any of these regex
+    '.*test_at.*',
     '.*test_funcs.*',
     '.*test_utils.*',
     '.*test_version.*',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dynamic = ["version"]
-dependencies = ["array-api-compat>=1.10.0"]
+dependencies = ["array-api-compat>=1.10.0,<2"]
 
 [project.optional-dependencies]
 tests = [
@@ -62,7 +62,7 @@ channels = ["https://prefix.dev/conda-forge"]
 platforms = ["linux-64", "osx-arm64", "win-64"]
 
 [tool.pixi.dependencies]
-python = ">=3.10.16,<3.14"
+python = ">=3.10,<3.14"
 array-api-compat = ">=1.10.0,<2"
 
 [tool.pixi.pypi-dependencies]
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -623,8 +623,10 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
 
         You may use two alternate syntaxes::
 
-          at(x, idx).set(value)  # or add(value), etc.
-          at(x)[idx].set(value)
+          >>> import array_api_extra as xpx
+          >>> xpx.at(x, idx).set(value)  # or add(value), etc.
+          >>> xpx.at(x)[idx].set(value)
+
     copy : bool, optional
         True (default)
             Ensure that the inputs are not modified.
@@ -647,14 +649,15 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
     (a) When you use ``copy=None``, you should always immediately overwrite
     the parameter array::
 
-        x = at(x, 0).set(2, copy=None)
+        >>> import array_api_extra as xpx
+        >>> x = xpx.at(x, 0).set(2, copy=None)
 
     The anti-pattern below must be avoided, as it will result in different
     behaviour on read-only versus writeable arrays::
 
-        x = xp.asarray([0, 0, 0])
-        y = at(x, 0).set(2, copy=None)
-        z = at(x, 1).set(3, copy=None)
+        >>> x = xp.asarray([0, 0, 0])
+        >>> y = xpx.at(x, 0).set(2, copy=None)
+        >>> z = xpx.at(x, 1).set(3, copy=None)
 
     In the above example, ``x == [0, 0, 0]``, ``y == [2, 0, 0]`` and z == ``[0, 3, 0]``
     when ``x`` is read-only, whereas ``x == y == z == [2, 3, 0]`` when ``x`` is
@@ -667,9 +670,10 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
 
         >>> import numpy as np
         >>> import jax.numpy as jnp
-        >>> at(np.asarray([123]), np.asarray([0, 0])).add(1)
+        >>> import array_api_extra as xpx
+        >>> xpx.at(np.asarray([123]), np.asarray([0, 0])).add(1)
         array([124])
-        >>> at(jnp.asarray([123]), jnp.asarray([0, 0])).add(1)
+        >>> xpx.at(jnp.asarray([123]), jnp.asarray([0, 0])).add(1)
         Array([125], dtype=int32)
 
     See Also
@@ -686,21 +690,22 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
     --------
     Given either of these equivalent expressions::
 
-      x = at(x)[1].add(2, copy=None)
-      x = at(x, 1).add(2, copy=None)
+      >>> import array_api_extra as xpx
+      >>> x = xpx.at(x)[1].add(2, copy=None)
+      >>> x = xpx.at(x, 1).add(2, copy=None)
 
     If x is a JAX array, they are the same as::
 
-      x = x.at[1].add(2)
+      >>> x = x.at[1].add(2)
 
     If x is a read-only numpy array, they are the same as::
 
-      x = x.copy()
-      x[1] += 2
+      >>> x = x.copy()
+      >>> x[1] += 2
 
     For other known backends, they are the same as::
 
-      x[1] += 2
+      >>> x[1] += 2
     """
 
     _x: Array
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -848,7 +848,11 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         xp: ModuleType | None = None,
     ) -> Array:  # numpydoc ignore=PR01,RT01
         """Apply ``x[idx] += y`` and return the updated array."""
-        return self._iop("add", operator.add, y, copy=copy, xp=xp)
+
+        # Note for this and all other methods based on _iop:
+        # operator.iadd and operator.add subtly differ in behaviour, as
+        # only iadd will trigger exceptions when y has an incompatible dtype.
+        return self._iop("add", operator.iadd, y, copy=copy, xp=xp)
 
     def subtract(
         self,
@@ -858,7 +862,7 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         xp: ModuleType | None = None,
     ) -> Array:  # numpydoc ignore=PR01,RT01
         """Apply ``x[idx] -= y`` and return the updated array."""
-        return self._iop("subtract", operator.sub, y, copy=copy, xp=xp)
+        return self._iop("subtract", operator.isub, y, copy=copy, xp=xp)
 
     def multiply(
         self,
@@ -868,7 +872,7 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         xp: ModuleType | None = None,
     ) -> Array:  # numpydoc ignore=PR01,RT01
         """Apply ``x[idx] *= y`` and return the updated array."""
-        return self._iop("multiply", operator.mul, y, copy=copy, xp=xp)
+        return self._iop("multiply", operator.imul, y, copy=copy, xp=xp)
 
     def divide(
         self,
@@ -878,7 +882,7 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         xp: ModuleType | None = None,
     ) -> Array:  # numpydoc ignore=PR01,RT01
         """Apply ``x[idx] /= y`` and return the updated array."""
-        return self._iop("divide", operator.truediv, y, copy=copy, xp=xp)
+        return self._iop("divide", operator.itruediv, y, copy=copy, xp=xp)
 
     def power(
         self,
@@ -888,7 +892,7 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         xp: ModuleType | None = None,
     ) -> Array:  # numpydoc ignore=PR01,RT01
         """Apply ``x[idx] **= y`` and return the updated array."""
-        return self._iop("power", operator.pow, y, copy=copy, xp=xp)
+        return self._iop("power", operator.ipow, y, copy=copy, xp=xp)
 
     def min(
         self,
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -1,12 +1,22 @@
 """Public API Functions."""
 
+import operator
 import warnings
 
+# https://github.com/pylint-dev/pylint/issues/10112
+from collections.abc import Callable  # pylint: disable=import-error
+from typing import ClassVar, Literal
+
 from ._lib import _compat, _utils
-from ._lib._compat import array_namespace
-from ._lib._typing import Array, ModuleType
+from ._lib._compat import (
+    array_namespace,
+    is_jax_array,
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -26,7 +26,11 @@ classifiers = [
     "Typing :: Typed",
 ]
 dynamic = ["version"]
-dependencies = ["array-api-compat>=1.1.1"]
+# dependencies = [ "array-api-compat" >= 1.1.1 ]  # DNM
+dependencies = [
+    # DNM main plus #211, #205, #207
+    "array-api-compat @ git+https://github.com/crusaderky/array-api-compat.git@9761bf2c2790416e3c117ed9f64407525b4a3498"
+]
```
crusaderky, line 33: This doesn't work and I could use the help of someone more skilled in pixi-fu.
I want to test this PR (locally and in CI) against my latest changes to array-api-compat, before they are pushed into a new release.
The updated pixi.lock looks correct (see above). However, when I run `pixi run -e tests tests` I get
```
E   ImportError: cannot import name 'is_writeable_array' from 'array_api_compat' (/home/crusaderky/github/array-api-extra/.pixi/envs/tests/lib/python3.13/site-packages/array_api_compat/__init__.py)
```
`is_writeable_array` is exactly the function that has been added in the git branch temporarily referenced above.
if I look at `.pixi/envs/tests`, however, I can see that pixi installed the latest released version (1.9.1) instead.

src/array_api_extra/_funcs.py
```diff
@@ -546,3 +552,275 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+            # __getitem__ with fancy index on a numpy array
+            writeable = True
+            copy = False
+        else:
+            raise ValueError(f"Invalid value for copy: {copy!r}")
+
+        if copy:
+            try:
+                at_ = x.at
+            except AttributeError:
```
crusaderky, line 716: The alternative to this duck-type pattern is to import `is_jax_array` from array-api-compat. @lucascolley please advise on what you'd prefer.

tests/test_at.py
```diff
@@ -0,0 +1,174 @@
...
+    is_jax_array,
+    is_pydata_sparse_array,
+    is_torch_array,
+    is_torch_namespace,
+    is_writeable_array,
+)
+
+from array_api_extra import at
+
+from ._helpers import all_libraries, import_
```
crusaderky, line 20: WIP still need to refactor this.

@lucascolley this test module necessarily needs to add all libraries (numpy, jax, dask, torch, sparse, cupy) to CI in order to test the functionality, whereas the rest of the test suite exclusively tests vs. array-api-strict.
Should I add these libraries to ci.yml or should I leave it alone and create a brand new github workflow?

tests/test_at.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def assert_array_equal(a, b):
+    if is_pydata_sparse_array(a):
+        a = a.todense()
+    elif is_cupy_array(a):
+        a = a.get()
+    elif is_dask_array(a):
+        a = a.compute()
+    np.testing.assert_array_equal(a, b)
```
lucascolley, line 30: See also gh-17

pyproject.toml
lucascolley: https://pixi.sh/latest/reference/pixi_manifest/#full-specification might help you, otherwise the pixi devs are super helpful on the prefix-dev discord

tests/test_at.py
lucascolley: Seems fine to add to the existing workflow for now. You can create a separate pixi `ci-` env with all the libraries

pyproject.toml
```diff
@@ -227,6 +235,9 @@ ignore = [
     "PLR09",   # Too many <...>
     "PLR2004", # Magic value used in comparison
     "ISC001",  # Conflicts with formatter
+    "EM101",   # raw-string-in-exception
```
lucascolley, line 238: I would rather fix these, the justification given on the rule page seems reasonable to me: https://docs.astral.sh/ruff/rules/raw-string-in-exception/

pyproject.toml
crusaderky: It was a bug https://github.com/prefix-dev/pixi/issues/2677

crusaderky:
- [x] Implementation
- [x] Tests pass locally
- [ ] Make linters happy
- [ ] CI
- [ ] Review docs

crusaderky:
Rebased on top of #58

crusaderky:
Any idea how to fix this? Looks like the ubuntu_latest VM has an obsolete driver

> ERROR cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,348 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+                i is not None and i is not Ellipsis and not isinstance(i, slice)
+                for i in tup_idx
+            ):
+                msg = "get() with a scalar index typically returns a copy"
+                raise ValueError(msg)
+
+            # Note: this is not the same list of backends as is_writeable_array()
+            if is_dask_array(x) or is_jax_array(x) or is_pydata_sparse_array(x):
+                msg = f"get() on {array_namespace(x)} arrays always returns a copy"
+                raise ValueError(msg)
```
crusaderky, line 728: These checks are very brittle and I'm not comfortable going on with `get()` as it is.

I can see two ways forward:
1. Write a new function in `array-api-compat`, `getitem_returns_view(x: Array, idx: Index) -> bool` that explores all the nooks and crannies of all the possible intersections of arrays and indices, then use it here
2. Remove `get()` altogether. I have a strong suspicion it may not be needed in scipy to begin with.

crusaderky: The only use I can think of out of `get(copy=True)` would be to avoid an unnecessary double copy when idx is a boolean mask

crusaderky:
> I don't think we can get GPU CI without paying someone for it, cc @rgommers .

Fixed by swapping `cupy` with `cupy-core`, which is CPU-only.

crusaderky:
> I don't think we can get GPU CI without paying someone for it, cc @rgommers .

Can cupy run on a CPU-only host?
I've disabled it in CI for the time being

rgommers:
> I don't think we can get GPU CI without paying someone for it, cc @rgommers .

Yep, that's on my radar to push forward this month, on multiple projects. Please feel free to open a new issue and assign it to me. I think we can hook up a shared GPU runner between this project and `array-api-compat`.

> Can cupy run on a CPU-only host?

I don't think so.

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    Warning
+    -------
+    The array API standard does not support integer array indices.
+    The behaviour of update methods when the index is an array of integers
+    is undefined; this is particularly true when the index contains multiple
+    occurrences of the same index, e.g. ``at(x, [0, 0]).set(2)``.
+
+    Note
+    ----
+    `sparse <https://sparse.pydata.org/>`_ is not supported by update methods yet.
```
lucascolley, line 648: can you also add a note that read only arrays from any array library not supported by array-api-compat are not supported here?

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+    See Also
+    --------
+    `jax.numpy.ndarray.at <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html>`_
+    """
+
+    x: Array
+    idx: Index
```
lucascolley, line 656: Currently these attributes render in the docs as

image

Could we fix that?

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+
+_undef = object()
+
+
+class at:  # pylint: disable=invalid-name
+    """
+    Update operations for read-only arrays.
+
+    This implements ``jax.numpy.ndarray.at`` for all backends.
```
lucascolley, line 573: Can add a clarification along the lines of my comment about read-only arrays from unknown libraries below.

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+    Parameters
+    ----------
+    x : array
+        Input array.
+    idx : index, optional
```
lucascolley, line 579: you could link to https://data-apis.org/array-api/latest/API_specification/indexing.html in the notes and perhaps reference it here, to explain what an `index` is

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+          at(x, idx).set(value)  # or get(), add(), etc.
+          at(x)[idx].set(value)
+
+    copy : bool, optional
+        True (default)
+            Ensure that the inputs are not modified.
+        False
+            Ensure that the update operation writes back to the input.
+            Raise ValueError if a copy cannot be avoided.
```
lucascolley, line 590:
```suggestion
            Raise ``ValueError`` if a copy cannot be avoided.
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    If x is a JAX array, they are the same as::
+
+      x = x.at[1].add(2)
+
+    If x is a read-only numpy array, they are the same as::
+
+      x = x.copy()
+      x[1] += 2
+
+    Otherwise, they are the same as::
```
lucascolley, line 618:
```suggestion
    For other known backends, they are the same as::
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+      x[1] += 2
+
+    Warning
+    -------
+    When you use copy=None, you should always immediately overwrite
```
lucascolley, line 624:
```suggestion
    When you use ``copy=None``, you should always immediately overwrite
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+    The anti-pattern below must be avoided, as it will result in different behaviour
+    on read-only versus writeable arrays::
+
+        x = xp.asarray([0, 0, 0])
+        y = at(x, 0).set(2, copy=None)
+        z = at(x, 1).set(3, copy=None)
+
+    In the above example, ``x == [0, 0, 0]``, ``y == [2, 0, 0]`` and z == ``[0, 3, 0]``
+    when x is read-only, whereas ``x == y == z == [2, 3, 0]`` when x is writeable!
```
lucascolley, line 637:
```suggestion
    In the above example, ``x == [0, 0, 0]``, ``y == [2, 0, 0]`` and ``z == [0, 3, 0]``
    when ``x`` is read-only, whereas ``x == y == z == [2, 3, 0]`` when ``x`` is writeable!
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +561,347 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+    Warning
+    -------
+    The array API standard does not support integer array indices.
+    The behaviour of update methods when the index is an array of integers
+    is undefined; this is particularly true when the index contains multiple
+    occurrences of the same index, e.g. ``at(x, [0, 0]).set(2)``.
```
lucascolley, line 644: for my understanding, can you clarify what this warning means, referencing the standard?

lucascolley (commented):
overall this looks great to me, thanks! Added some comments.

I'll take another look here before it goes in. In the meantime, hopefully the upstream work goes smoothly, and it would be fantastic if @jakevdp / @rgommers  / anyone else could take a closer look at the implementation!

pyproject.toml
```diff
@@ -225,6 +262,7 @@ ignore = [
     "PLR09",   # Too many <...>
     "PLR2004", # Magic value used in comparison
     "ISC001",  # Conflicts with formatter
+    "PD008",   # Use `.loc` instead of `.at`
```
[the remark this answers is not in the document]
crusaderky: Everywhere. The linter seems to think that any `Any` object with `.at` is a pandas dataframe.

crusaderky:
I gave a final round of polish. This is ready for final review and approval.
Note the [blocker PRs and DNMs](https://github.com/data-apis/array-api-extra/pull/53#issue-2727108419) before it can be merged though.

crusaderky:
> > I don't think we can get GPU CI without paying someone for it, cc @rgommers .
> 
> Yep, that's on my radar to push forward this month, on multiple projects. Please feel free to open a new issue and assign it to me. I think we can hook up a shared GPU runner between this project and `array-api-compat`.
> 
> > Can cupy run on a CPU-only host?
> 
> I don't think so.

https://github.com/data-apis/array-api-extra/issues/60

src/array_api_extra/_funcs.py
```diff
@@ -1,10 +1,22 @@
+from __future__ import annotations
```
crusaderky, line 1: Needed due to `__getitem__(self, idx) -> at:`

pyproject.toml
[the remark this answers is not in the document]
lucascolley: FWIW I can't reproduce the failures locally - do you have a MWE I could give to the ruff devs?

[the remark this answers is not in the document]
crusaderky: After I did `rm -rf .pixi pixi.lock`, I cannot reproduce this either anymore. I'm puzzled.

src/array_api_extra/_funcs.py
lucascolley: I think you can remove it if you change that to `__getitem__(self, idx) -> "at":`.

When we drop Python 3.10 we can then change it to `-> Self`.

pyproject.toml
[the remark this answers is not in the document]
lucascolley: ruff put out a release 1 hour ago so maybe they fixed a bug 🤷‍♂️

src/array_api_extra/_funcs.py
crusaderky: I've removed get(), at least for the time being.

lucascolley:
I resolved the merge conflicts after adding `numpydoc` to pre-commit and made some minor tweaks to the docs.

crusaderky:
@lucascolley if you don't mind building against array-api-compat git tip until their next release, this PR is ready for final review and merge

pyproject.toml
```diff
@@ -99,7 +101,9 @@ tests-cov = "pytest -v -ra --cov --cov-report=xml --cov-report=term --durations=
 
 clean-vendor-compat = "rm -rf vendor_tests/array_api_compat"
 clean-vendor-extra = "rm -rf vendor_tests/array_api_extra"
-copy-vendor-compat = { cmd = "cp -r $(python -c 'import site; print(site.getsitepackages()[0])')/array_api_compat vendor_tests/", depends-on = ["clean-vendor-compat"] }
+copy-vendor-compat = { cmd = "cp -r $(python -c 'import site; print(site.getsitepackages()[0])')/array_api_compat vendor_tests/", depends-on = [
+    "clean-vendor-compat",
+] }
```
lucascolley, line 106: this might have been from one of my commits due to an IDE extension, but we should keep this on one line

lucascolley (commented):
> @lucascolley if you don't mind building against array-api-compat git tip until their next release, this PR is ready for final review and merge

That's fine with me, but might be cleaner to get an array-api-compat release out first unless there are any big blockers.

Let's give @rgommers and @jakevdp time to take a look if they would like to.

Once (or before) this is merged, do you think you could make a PR to my branch at https://github.com/scikit-learn/scikit-learn/pull/30340? I think we will have to transition scikit-learn from using array-api-compat as an optional dependency to vendoring it, but that sounds feasible based on https://github.com/scikit-learn/scikit-learn/issues/30367#issuecomment-2506749233/.

lucascolley (approved):

crusaderky:
> Once (or before) this is merged, do you think you could make a PR to my branch at [scikit-learn/scikit-learn#30340](https://github.com/scikit-learn/scikit-learn/pull/30340)? I think we will have to transition scikit-learn from using array-api-compat as an optional dependency to vendoring it, but that sounds feasible based on [scikit-learn/scikit-learn#30367 (comment)](https://github.com/scikit-learn/scikit-learn/issues/30367#issuecomment-2506749233).

https://github.com/lucascolley/scikit-learn/pull/2

src/array_api_extra/_funcs.py
```diff
@@ -538,3 +548,314 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        Allow for the alternate syntax ``at(x)[start:stop:step]``.
+
+        It looks prettier than ``at(x, slice(start, stop, step))``
+        and feels more intuitive coming from the JAX documentation.
+        """
+        if self._idx is not _undef:
+            msg = "Index has already been set"
+            raise ValueError(msg)
+        self._idx = idx
+        return self
```
jakevdp, line 676: Maybe it would make sense to return a shallow copy rather than mutating `self`. Otherwise, someone might write something like this and be surprised by the behavior:
```
getter = at(x)
y = getter[0].add(1)
z = getter[1].add(2)
```

src/array_api_extra/_funcs.py
```diff
@@ -538,3 +548,314 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        Note that this is not the same as
+            operator.iadd(x[idx], y)
+        Consider for example when x is a numpy array and idx is a fancy index, which
+        triggers a deep copy on __getitem__.
+        """
+        res, x = self._update_common(at_op, y, copy=copy, xp=xp)
+        if res is not None:
+            return res
+        assert x is not None
+        x[self._idx] = elwise_op(x[self._idx], y)
```
jakevdp, line 784: As currently implemented, the `_iop` path has different semantics for repeated indices between JAX and NumPy:
```python
>>> import numpy as np
>>> x = np.zeros(4)
>>> idx = np.array([1, 2, 2, 3, 3, 3])
>>> x[idx] = x[idx] + 1
>>> x
array([0., 1., 1., 1.])

>>> import jax.numpy as jnp
>>> x = jnp.zeros(4)
>>> idx = jnp.array([1, 2, 2, 3, 3, 3])
>>> x.at[idx].add(1)
Array([0., 1., 2., 3.], dtype=float32)
```
At the very least, the difference should be documented.

jakevdp (changes requested):

src/array_api_extra/_funcs.py
crusaderky: it's already documented at lines 612:622

src/array_api_extra/_funcs.py
crusaderky: Good idea, done

lucascolley (approved):

pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 dynamic = ["version"]
-dependencies = ["array-api-compat>=1.1.1"]
+dependencies = ["array-api-compat>=1.10.0,<2"]
```
crusaderky, line 29: @lucascolley Updated to match tool.pixi.dependencies

pyproject.toml
```diff
@@ -62,8 +62,8 @@ channels = ["https://prefix.dev/conda-forge"]
 platforms = ["linux-64", "osx-arm64", "win-64"]
 
 [tool.pixi.dependencies]
-python = ">=3.10.15,<3.14"
-array-api-compat = ">=1.1.1"
+python = ">=3.10,<3.14"
```
crusaderky, line 65: Was there any purpose in specifying the patch version of Python?

crusaderky:
@lucascolley ready for merge and release! :partying_face:

lucascolley:
good to proceed @rgommers ?

src/array_api_extra/_funcs.py
```diff
@@ -589,3 +599,313 @@ def pad(
...
+    `sparse <https://sparse.pydata.org/>`_, as well as read-only arrays from libraries
+    not explicitly covered by ``array-api-compat``, are not supported by update
+    methods.
+
+    Examples
+    --------
+    Given either of these equivalent expressions::
+
+      x = at(x)[1].add(2, copy=None)
+      x = at(x, 1).add(2, copy=None)
```
lucascolley, line 690: this doesn't have to land in this PR, but I think it would be good to use `import array_api_extra as xpx` ... `x = xpx.at(...)` as that is what is used in the other examples in the docs

crusaderky: Done; let me know if it's what you meant

lucascolley:
looks great, thanks! Note to self: we ought to add

```
# sphinx-copybutton configurations
copybutton_prompt_text = r">>> |\.\.\. |\$ |In \[\d*\]: | {2,5}\.{3,}: | {5,8}: "
copybutton_prompt_is_regexp = True
```

to `docs/conf.py`.

crusaderky:
Annoyingly, codecov is glitching (again). It's stating in the "Files" view that the tests never run under JAX, but if you follow the [link from Check ci-backends](https://app.codecov.io/github/data-apis/array-api-extra/commit/98141cae091c4cea2cf6df7f4f2eefd2eb235605/blob/src/array_api_extra/_funcs.py) it says that everything's green.

lucascolley:
yeah, happy to just ignore that, but also happy if anyone figures out how to fix that

rgommers (commented):
Thanks @crusaderky and reviewers. This is looking quite good. A couple of comments (disclaimer: it's a little hard to dig through the already-resolved review comments, so I may be saying something that was already discussed there):

(1) The `copy=True` default seems wrong to me. The main point of this feature is that we can replace in-place operations like `x[idx, :] += y` in SciPy et al. with something new to support JAX. That something new should be the idiomatic JAX way, so `x = xpx.at(x)[idx, :].add(y)`. This should come with as little of a performance penalty for NumPy usage as possible - hence the `copy=True` default doesn't work. It's already a significant concession to code readability to have to avoid `+=` & co, and having to also add `copy=None` in all places where performance matters seems too much.

(2) When playing around with this branch, most things work as expected, but this was a little surprising:
```python
>>> x = np.arange(5)
>>> x = xpx.at(x)[[3, 4]].divide(1.5)
>>> x.dtype
dtype('int64')
>>> x = np.arange(5)
>>> x /= 1.5
Traceback (most recent call last):
  Cell In[32], line 1
    x /= 1.5
UFuncTypeError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```
Is this on purpose or by accident? Raising on these cases may be cleaner if that can be done in a performant way.

(3) Library support is pretty comprehensive already; a few others come to mind. MLX is already supported by the generic path since [it allows in-place syntax](https://ml-explore.github.io/mlx/build/html/usage/indexing.html#in-place-updates) (I think, didn't verify). It also [has array.at](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.array.at.html#mlx.core.array.at) which matches JAX syntax/semantics. `ndonnx` is going to need explicit support in the future as well, since it's immutable like JAX, but that can be left for later since it doesn't seem to have `Array.at` or another alternative here. tl;dr nothing to change in this PR I think.

src/array_api_extra/_funcs.py
```diff
@@ -589,3 +599,318 @@ def pad(
...
+    ) -> None:  # numpydoc ignore=GL08
+        self._x = x
+        self._idx = idx
+
+    def __getitem__(self, idx: Index, /) -> "at":  # numpydoc ignore=PR01,RT01
+        """
+        Allow for the alternate syntax ``at(x)[start:stop:step]``.
+
+        It looks prettier than ``at(x, slice(start, stop, step))``
+        and feels more intuitive coming from the JAX documentation.
```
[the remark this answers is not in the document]
crusaderky: I've been writing several prototype PRs on various submodules of Sphinx.
From the experience I gathered there, I must say that `at(x)[idx]` feels more natural and easier on the eyes when there are multiple indices and/or slies, whereas `at(x, idx)` feels nicer and less cluttered in all other cases.

crusaderky:
> (1) The copy=True default seems wrong to me.

Changed it to copy=None

> (2) x /= 1.5

TIL that 

- ``x[idx] = x[idx].__add__(value)``
- ``x[idx] = x[idx].__iadd__(value)``

are _almost_ always equivalent!

Fixed.

src/array_api_extra/_funcs.py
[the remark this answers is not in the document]
rgommers: Okay, fair enough - seems like an argument to keep both then.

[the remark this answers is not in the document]
crusaderky: In fact, I've been considered writing aliases for the benefit of readability `set_at(x, idx, value)` etc. but haven't gone for them for the sake of keeping the API surface small.

tests/test_at.py
```diff
@@ -0,0 +1,158 @@
...
+
+    with pytest.raises(ValueError, match="Index"):
+        at(a).set(4)
+    with pytest.raises(ValueError, match="Index"):
+        at(a, 0)[0].set(4)
+
+
+@pytest.mark.parametrize("copy", [True, False])
+@pytest.mark.parametrize("op", ["add", "subtract", "multiply", "divide", "power"])
+def test_iops_incompatible_dtype(op, copy):
```
lucascolley, line 143:
```suggestion
def test_iops_incompatible_dtype(op: str, copy: bool):
```

lucascolley:
shall we get going with a release? :)

src/array_api_extra/_funcs.py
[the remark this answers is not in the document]
rgommers: Not adding a third way sounds like a healthy idea at least for now - let's see how we like this in its current form I'd say.

rgommers (approved):
Looks ready to me now, +1 for merge & release if everyone else is happy too.

lucascolley:
let's give this a whirl, thanks all! if you can point the SciPy PR to the release when it is out @crusaderky, I can update the sklearn PR

PR #58, opened 2024-12-10
Saner linters

crusaderky:
Thorough review of the linters in use. This was triggered by #53, where I found many unresolvable conflicts among the linters.


.github/workflows/ci.yml
```diff
@@ -35,11 +35,8 @@ jobs:
           pixi-version: v0.39.0
           cache: true
           environments: lint
-      - name: Run Pylint, Mypy & Pyright
-        run: |
-          pixi run -e lint pylint
```
crusaderky, line 40: Redundant and incompatible with ruff

.github/workflows/ci.yml
```diff
@@ -35,11 +35,8 @@ jobs:
...
-          pixi run -e lint mypy
-          pixi run -e lint pyright
```
crusaderky, line 42: Can't have pyright AND mypy in the same project. They _will_ eventually result conflicting with each other, and in fact they did so heavily in #53.
Chose mypy as I found pyright extremely obnoxious and factually wrong in too many cases.

pyproject.toml
```diff
@@ -70,9 +70,7 @@ array-api-extra = { path = ".", editable = true }
 
 [tool.pixi.feature.lint.dependencies]
 pre-commit = "*"
-pylint = "*"
-basedmypy = "*"
```
crusaderky, line 74: replaced basedmypy with mypy. I think it is an extremely bad idea to diverge from PEPs. Namely, IDE plugins will not support this.

lucascolley:
Overall I'm -1 on this change. I'm happy to investigate conflicts.

> Can't have pyright AND mypy in the same project.

That isn't true, it is done successfully in @jorenham's scipy-stubs.

rgommers:
Quick comment: conceptually this PR seems right. Best to pick 1 linter and 1 type checker, everything else is pretty much unworkable once the code starts getting more complex. I'd personally pick `ruff` (for autofix and well-maintained) and `pyright` (much better maintained and less buggy than mypy), but it doesn't really matter what the choices are as long as there's only one tool per type of check.

pyproject.toml
jorenham: This is untrue. All mypy IDE plugins and other CI tools also work with `basedmypy`. 

In fact, mypy diverges more from the typing specification that `basedmypy` does. It also has [significantly more bugs](https://github.com/erictraut/mypy_issues).

.github/workflows/ci.yml
jorenham: > Can't have pyright AND mypy in the same project. They will eventually result conflicting with each other, and in fact they did so heavily in https://github.com/data-apis/array-api-extra/pull/53.

This isn't true at all, and all of my typed projects are proof of this.
In my experience, there are indeed situations where mypy and pyright behave differently. In at least 99% of these cases, this is caused by one of the >1200 confirmed bugs in mypy that are present in pyright: https://github.com/erictraut/mypy_issues

> Chose mypy as I found pyright extremely obnoxious and factually wrong in too many cases.

Do you have an example of this?

.github/workflows/ci.yml
jorenham: Pylint is also included in ruff, so why would it be incompatible?

jorenham:
> Best to pick 1 linter and type checker

In case of linters I'd agree with this, unless they're actually linting different things. In my project I always use (only) ruff for python, but I also always markdownlint, and in most cases also [repo-review](https://github.com/scientific-python/repo-review). These are all disjoint, so I see no problem with that.

I don't agree that you should only should pick one type checker. I always use both mypy and pyright in my projects, because I want the users to also be able to use both, which requires me to verify that both type-checkers actually understand the typing annotations.
Mypy and pyright certainly overlap, but also differ in many situations. This is mostly because mypy has >1200 bugs that pyright doesn't have ([source](https://github.com/erictraut/mypy_issues)). But unfortunately, more than half of the users use mypy at the moment, so I can't just ignore it.

lucascolley:
RE: pylint, [the Scientific Python Development guide says](https://learn.scientific-python.org/development/guides/style/#pylint-noisy)

> PyLint is very opinionated, with a high signal-to-noise ratio. However, by limiting the default checks or by starting off a new project using them, you can get some very nice linting, including catching some problematic code that otherwise is hard to catch.

We currently use the recommended limits on the default checks. If there are conflicts with ruff then it shouldn't be difficult to extend that list.

.github/workflows/ci.yml
KotlinIsland: there are many pylint rules that ruff hasn't implemented: https://github.com/astral-sh/ruff/issues/970

KotlinIsland:
typeshed uses both mypy and pyright

lucascolley (changes requested):
it looks like this is workable without changing the existing config - see gh-59

.github/workflows/ci.yml
crusaderky: A simple example was 
```python
class at:
```
pylint complains that classes need to be capitalized;
but if you add a `# noqa` to silene pylint, the ruff fails with an "unnecessary noqa" error

lucascolley: the way to silence pylint is `  # pylint: disable=invalid-name`

PR #59, opened 2024-12-10
POC: appease linter for gh-53

lucascolley:
Looks like there are no unresolvable or essential conflicts here, just required a little bit of wrangling :)

@jorenham would you be able to review the typing? I've added a few `Any` aliases where I wasn't sure how to annotate.

@crusaderky once we check that the linter is happy here, shall I make a PR to your branch? Then you can continue your work in gh-53.


src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        elif copy is False:
+            writeable = is_writeable_array(x)
+            if not writeable:
+                msg = "Cannot modify parameter in place"
+                raise ValueError(msg)
+        elif copy is None:  # type: ignore[redundant-expr]
+            writeable = is_writeable_array(x)
+            copy = _is_update and not writeable
+        else:
+            msg = f"Invalid value for copy: {copy!r}"  # type: ignore[unreachable]  # pyright: ignore[reportUnreachable]
```
lucascolley, line 718: is there a better way to deal with situations like this where invalid types can be passed at runtime?

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    See Also
+    --------
+    `jax.numpy.ndarray.at <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html>`_
+    """
+
+    x: Array
+    idx: Index
+    __slots__: ClassVar[tuple[str, str]] = ("idx", "x")
+
+    def __init__(self, x: Array, idx: Index = _undef, /):
```
jorenham, line 662: mypy strict mode requires this I believe

```suggestion
    def __init__(self, x: Array, idx: Index = _undef, /) -> None:
```

src/array_api_extra/_lib/_typing.py
```diff
@@ -10,6 +10,8 @@
     # To be changed to a Protocol later (see data-apis/array-api#589)
     Array = Any  # type: ignore[no-any-explicit]
     Device = Any  # type: ignore[no-any-explicit]
+    Index = Any  # type: ignore[no-any-explicit]
+    Untyped = Any  # type: ignore[no-any-explicit]
```
jorenham, line 14: how how this?

```suggestion
    Untyped = Any  # type: ignore[no-any-explicit]
    Array = Untyped
    Device = Untyped
    Index = Untyped
```

tests/test_at.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        try:
+            lib = import_module(library)
+        except ImportError:
+            pytest.skip(f"{library} is not installed")
+        x = lib.asarray([10.0, 20.0, 30.0])
+    return x
+
+
+def assert_array_equal(a: Array, b: Array) -> None:
```
lucascolley, line 48: `None` or `NoReturn`?

src/array_api_extra/_lib/_typing.py
```diff
@@ -19,4 +21,4 @@ def no_op_decorator(f):  # pyright: ignore[reportUnreachable]
 
 __all__ = ["ModuleType", "override"]
 if typing.TYPE_CHECKING:
-    __all__ += ["Array", "Device"]
+    __all__ += ["Array", "Device", "Index", "Untyped"]
```
jorenham, line 24: I see not advantage of putting this in a `TYPE_CHECKING` block. And I can imagine that mypy wouldn't know how to deal with this, but I might be wrong here.

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
         xp.asarray(xp.finfo(x.dtype).eps, dtype=x.dtype, device=x.device),
     )
     return xp.sin(y) / y
+
+
+_undef = object()
+
+
+class at:  # pylint: disable=invalid-name
```
jorenham, line 567: I'm not sure if this is what you want, but it'll make the methods that return `-> at`  instead of `-> Self` theoretically valid.

```suggestion
@typing.final
class at:  # pylint: disable=invalid-name
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        return self
+
+    def _common(
+        self,
+        at_op: str,
+        y: Array = _undef,
+        /,
+        copy: bool | None = True,
+        xp: ModuleType | None = None,
+        _is_update: bool = True,
```
jorenham, line 684: or perhaps

```suggestion
        *,
        _is_update: bool = True,
```

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        **kwargs: Untyped,
+    ) -> tuple[Untyped, None] | tuple[None, Array]:
```
jorenham, line 686: is it perhaps possible to `@overload` these cases?

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+            if writeable is None:
+                writeable = is_writeable_array(x)
+            if not writeable:
+                # sparse crashes here
+                msg = f"Array {x} has no `at` method and is read-only"
+                raise ValueError(msg)
+
+        return None, x
+
+    def get(self, **kwargs: Untyped) -> Untyped:
```
jorenham, line 749:
```suggestion
    def get(self, /, **kwargs: Untyped) -> Untyped:
```

jorenham: given the `kwargs.get("copy")`, you at least know that `*, copy: bool | None = None`, right?

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        res, x = self._common("set", y, **kwargs)
+        if res is not None:
+            return res
+        assert x is not None
+        x[self.idx] = y
+        return x
+
+    def _iop(
+        self,
+        at_op: str,
```
jorenham, line 785: you could also consider using a `Literal["add", ...]` here

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        elwise_op: Callable[[Array, Array], Array],
```
jorenham, line 786: since `operator.add` is passed here, maybe this shouldn't be `(Array, Array) -> Array`, but something like `(Scalar, Scalar) -> Scalar` 🤔

tests/test_at.py
crusaderky: None. NoReturn is a function that either always raises or with an unbreakable infinite loop.

src/array_api_extra/_lib/_compat.pyi
```diff
@@ -11,3 +11,6 @@ def array_namespace(
     use_compat: bool | None = None,
 ) -> ArrayModule: ...
 def device(x: Array, /) -> Device: ...
+def is_array_api_obj(x: object, /) -> bool: ...
+def is_dask_array(x: object, /) -> bool: ...
+def is_writeable_array(x: object, /) -> bool: ...
```
jorenham, line 16: Maybe a `TypeGuard` could be used here.
I'm not sure if at this point it'll actually help, but it should at least be valid 🤷🏻

tests/test_at.py
```diff
@@ -0,0 +1,155 @@
...
+    is_pydata_sparse_array,
+    is_writeable_array,
+)
+
+from array_api_extra import at
+
+if TYPE_CHECKING:
+    from array_api_extra._lib._typing import Array, Untyped
+
+all_libraries = (
```
jorenham, line 21: `all_libraries: Final` will help typecheckers understand that this is a constant, which they usually automatically assume if written in all-caps

src/array_api_extra/_funcs.py
crusaderky: It isn't because the return type depends on a non-type test on x. And I'm definitely unwilling to explore writing a `class HasAtMethod(Protocol)` for a small internal function that is consumed exclusively 2 paragraph below.

src/array_api_extra/_funcs.py
crusaderky: overkill for a small internal method

src/array_api_extra/_funcs.py
crusaderky: Yes but I wanted the signature to be consistent with the other methods in the sphinx doc.
I'll probably change all methods to have an explicit `copy` and `xp` parameters.

src/array_api_extra/_funcs.py
crusaderky: x and y are always Array and this is a well-contained internal method; no need to make it as generic as possible

lucascolley:
broadly, I agree on not going overkill on the types. Let's just get something in that works.

src/array_api_extra/_funcs.py
jorenham: I'm not talking about making it generic 🤔

src/array_api_extra/_funcs.py
jorenham: I can see how you don't feel like putting in the effort to do this, but there are no downsides 🤷🏻‍♂️

src/array_api_extra/_funcs.py
```diff
@@ -548,3 +559,279 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    `sparse <https://sparse.pydata.org/>`_ is not supported by update methods yet.
+
+    See Also
+    --------
+    `jax.numpy.ndarray.at <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html>`_
+    """
+
+    x: Array
+    idx: Index
+    __slots__: ClassVar[tuple[str, str]] = ("idx", "x")
```
crusaderky, line 660: IMHO the linter should not force me to define the type of `__slots__`, because it's part of the python data model. This only adds attrition and reduces readability.

src/array_api_extra/_funcs.py
jorenham: > I'll probably change all methods to have an explicit `copy` and `xp` parameters.

Once you do, it could help avoid duplication by using a `TypedDict` and `Unpack` (which requires `typing_extensions` on `python < 3.11`) in that case.

src/array_api_extra/_funcs.py
crusaderky: This double-ignore is a strong example of why I'm strongly in favour of having only one type checker. This is tedious to both read and write.

src/array_api_extra/_funcs.py
jorenham: I agree

crusaderky:
> @crusaderky once we check that the linter is happy here, shall I make a PR to your branch?

No need I can cherry-pick

src/array_api_extra/_lib/_compat.pyi
crusaderky: TypeGuard is a very good idea, but I'd rather invest the energy to implement it in `array_api_compat`

src/array_api_extra/_funcs.py
lucascolley: let's punt on this for now, once we reach Python 3.11 there will be no need for `-> at`

src/array_api_extra/_funcs.py
jorenham: it's definitely annoying, but unfortunately it's the only possibility for ensuring compatibility with these two main type-checkers

jorenham: but I agree with mypy and pyright that the `else` clause here is redundant

jorenham: you could alternatively move the `if copy is None` to the top, and then do `elif copy: ... else: ...`

src/array_api_extra/_funcs.py
lucascolley: making this change - could you clarify how to use `_CanAt` @jorenham ?

src/array_api_extra/_funcs.py
jorenham: well, anything works with `operator.add` apparently 😅 

https://github.com/python/typeshed/blob/b0768596d2f7b2c12aa61d8a9714325284410b4b/stdlib/_operator.pyi#L55

src/array_api_extra/_funcs.py
lucascolley: what if we still want to throw an error at runtime if `copy="foo"` is passed? I guess the argument is that, since this is internal, that would always be caught by the static analysis anyway?

src/array_api_extra/_funcs.py
lucascolley: looks like this is no longer needed

lucascolley:
Feel free to cherry-pick from this point @crusaderky and we can continue in your PR

Issue #69, opened 2024-12-19
ENH: add `pad`


ev-br:
`np.pad` is not in array API spec, and is unlikely to get there. Still is a sometimes useful function, which is available in some backends (numpy, cupy, jax.numpy) but not others (torch). Thus it'd be great to add it to `array-api-extra`.

Implementing the full set of mode keywords is somewhat tricky, but the most useful one, `mode="constant"` is easy to implement even with pytorch. A ready implementation is available in the scipy PR:
https://github.com/scipy/scipy/pull/22122/files#diff-351836adc98d076c1552d17a57c52e6aa8ca43760bae44bea9190e55b4769b7fR873

The torch implementation derives from `torch._numpy`, https://github.com/pytorch/pytorch/blob/main/torch/_numpy/_funcs_impl.py#L2045

The version from the scipy PR is under the fold.

<details>

```
def xp_pad(x, pad_width, mode='constant', *, xp, **kwargs):
    # xp.pad is available on numpy, cupy and jax.numpy; on torch, reuse
    # http://github.com/pytorch/pytorch/blob/main/torch/_numpy/_funcs_impl.py#L2045
    # for mode = 'constant'
    if mode != 'constant':
        raise NotImplementedError()

    value = kwargs.get("constant_values", 0)

    if is_array_api_strict(xp):
        np_x = np.asarray(x)
        padded = np.pad(np_x, pad_width, mode=mode, **kwargs)
        return xp.asarray(padded)
    elif is_torch(xp):
        pad_width = xp.asarray(pad_width)
        pad_width = xp.broadcast_to(pad_width, (x.ndim, 2))
        pad_width = xp.flip(pad_width, axis=(0,)).flatten()
        return xp.nn.functional.pad(x, tuple(pad_width), value=value)
    else:
        return xp.pad(x, pad_width, mode=mode, **kwargs)
```

</details>

ev-br:
Closing as completed feels a bit premature.

PR #71, opened 2024-12-26
ENH: add `pad`

ev-br:
closes gh-69

Add an `np.pad` replacement, as discussed at https://github.com/scipy/scipy/pull/22122/files#r1895558533

### DIFF
--- a/src/array_api_extra/__init__.py
+++ b/src/array_api_extra/__init__.py
@@ -1,6 +1,6 @@
 """Extra array functions built on top of the array API standard."""
 
-from ._funcs import atleast_nd, cov, create_diagonal, expand_dims, kron, setdiff1d, sinc
+from ._funcs import atleast_nd, cov, create_diagonal, expand_dims, kron, setdiff1d, sinc, pad
 
 __version__ = "0.4.1.dev0"
 
@@ -14,4 +14,5 @@ __all__ = [
     "kron",
     "setdiff1d",
     "sinc",
+    "pad",
 ]
--- a/src/array_api_extra/__init__.py
+++ b/src/array_api_extra/__init__.py
@@ -1,6 +1,15 @@
 """Extra array functions built on top of the array API standard."""
 
-from ._funcs import atleast_nd, cov, create_diagonal, expand_dims, kron, setdiff1d, sinc, pad
+from ._funcs import (
+    atleast_nd,
+    cov,
+    create_diagonal,
+    expand_dims,
+    kron,
+    pad,
+    setdiff1d,
+    sinc,
+)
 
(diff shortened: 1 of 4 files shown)


lucascolley (commented):
Could you just add the array-agnostic path first in this PR? If we want to start adding dispatching to more efficient implementations for some libraries, we can consider adding that in a follow-up. I think I would rather dispatching code lived in a separate file to avoid cluttering the implementations.

There shouldn't be any need to use `is_array_api_strict_namespace` anywhere in this library.

ev-br:
- np.pad is a thing
- jax.numpy.pad is a thing
- torch.nn.functional.pad is a thing
- array_api_strict.pad is not
- "generic" implementation needs mutable arrays, so is not going to work on jax.numpy

I don't understnd the request, and I though we covered this in the scipy issue.

lucascolley:
By the way, if you notice any improvements that we should make to https://data-apis.org/array-api-extra/contributing.html, would be great to hear.

ev-br:
> I think we should still just add the "generic" implementation here, then look at dispatching immediately after that.

I still don't follow. This PR is I believe what scipy needs. What is wrong with the PR, what is the request exactly?

lucascolley:
> What is wrong with the PR, what is the request exactly?

I'll push changes to show you what I mean. Currently, the other functions in this library do not use any delegation to specific namespaces. Since we want to add that in order to use this in SciPy, this PR is doing 2 things which can be quite cleanly split:
1) add a `pad` which is similar to all of the other existing functions - array-agnostic, tested against array-api-strict.
2) add delegation to existing libraries on top, perhaps also test with those libraries

My request was just to do (1) first here before we think about how delegation should work in array-api-extra.

ev-br:
Ah, so the request is to rewrite the history into two commits: one for the "generic" implementation and the other with branching on namespace? This I can do easily. Or, like you said, feel free to rewrite the history here to your liking.

ev-br:
> I think there is potential to be smarter about how we do delegation, similar to what you and Matt have implemented in SciPy, if we are going to open up to the possibility of delegating for all functions in this library.

Okay.... I'd expect something like that is an overkill for a handful of if-elif-else branches in pure python code, but this is well in the "tastes differ" territory. Am rather curious to see what you have in mind.

lucascolley (approved):
this LGTM now if you're happy with it! We can look at adding delegation on top next.

ev-br (commented):
LGTM as an intermediate step, given that it's not usable on jax and suboptimal on other backends.
If we're adding more immediately after, updating the docs can be done in that immediately after.

PR #77, opened 2025-01-03
REL: set version to 0.5.0

lucascolley:


lucascolley:
@jorenham @kotlinisland I am seeing some `error: "object" not callable  [operator]` errors for objects which I'm pretty sure are in fact callable. This is with basedmypy 2.9.0. Any ideas?

lucascolley:
basedpyright reports no failures

lucascolley:
baedmypy 2.8.1 reports no failures

lucascolley:
I'll probably just pin basedmypy for this release, we can figure out what's going on after that

jorenham:
I believe this has to do with @KotlinIsland 's change in the bundled typeshed so that `builtins.getattr` now returns `object` instead of `Any`

KotlinIsland:
perhaps we could introduce a `no_strict_stdlib` config option to basedmypy? from experience these stricter types are very beneficial, but I understand that it doesn't fit every usecase

lucascolley:
it may well be that we are relying on some undefined behaviour, or have incorrectly typed something, so I'd rather check that before changing the config

KotlinIsland:
the issue is that the behavior of getting attrs is underspecified within the typesystem, even for the most trivial of cases:
```py
class A:
    a = 1
getattr(A, 'a')  # Any
```

lucascolley:
right, so what would be the advisable way to deal with this downstream? `type: ignore` comments? Change a config setting? add some extra annotations?

jorenham:
> right, so what would be the advisable way to deal with this downstream? `type: ignore` comments? Change a config setting? add some extra annotations?

`typing.cast` would be the cleanest in this situation I think

lucascolley:
PR welcome I guess 😉 (no pressure)

KotlinIsland:
sorry, what is your preferred action here?

> right, so what would be the advisable way to deal with this downstream?

if by downstream you are refering to this project, then i recommend writing your own `getattr` function that is explicitly typed to return `Any`. the motivation for this change originally was "the behavior of `getattr` should be typed by default, if you want it as `Any` then implement that yourself"

an alternative, that is more safe would be to cast each and every attr that is gotten to the expected type. i prefer this approach as it is more descriptive and type safe than using `Any`

lucascolley:
the cast approach looks fine to me!

PR #82, opened 2025-01-07
ENH: allow tuple pad_width in `pad`

ev-br:
Observed in https://github.com/scipy/scipy/pull/22226 : allow `pad_width` to be a tuple.

### DIFF
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -555,7 +555,7 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
 
 def pad(
     x: Array,
-    pad_width: int,
+    pad_width: int | tuple[int, int] | list[tuple[int, int]],
     mode: str = "constant",
     *,
     xp: ModuleType | None = None,
@@ -568,8 +568,12 @@ def pad(
     ----------
     x : array
         Input array.
-    pad_width : int
+    pad_width : int or tuple of ints or list of pairs of ints
         Pad the input array with this many elements from each side.
+        If a list of tuples, ``[(before_0, after_0), ... (before_N, after_N)]``,
+        each pair applies to the corresponding axis of ``x``.
+        A single tuple, ``(before, after)``, is equivalent to a list of ``x.ndim``
+        copies of this tuple.
     mode : str, optional
         Only "constant" mode is currently supported, which pads with
         the value passed to `constant_values`.
@@ -590,16 +594,43 @@ def pad(
 
     value = constant_values
 
+    # make pad_width a list of length-2 tuples of ints
+    if isinstance(pad_width, int):
+        pad_width = [(pad_width, pad_width)] * x.ndim
+
+    if isinstance(pad_width, tuple):
+        pad_width = [pad_width] * x.ndim
+
     if xp is None:
         xp = array_namespace(x)
 
+    slices = []
+    newshape = []
+    for ax, w_tpl in enumerate(pad_width):
+        if len(w_tpl) != 2:
+            msg = f"expect a 2-tuple (before, after), got {w_tpl}."
+            raise ValueError(msg)
+
+        sh = x.shape[ax]
+        if w_tpl[0] == 0 and w_tpl[1] == 0:
+            sl = slice(None, None, None)
+        else:
+            start, stop = w_tpl
+            stop = None if stop == 0 else -stop
+
+            sl = slice(start, stop, None)
+            sh += w_tpl[0] + w_tpl[1]
+
+        newshape.append(sh)
+        slices.append(sl)
+
     padded = xp.full(
-        tuple(x + 2 * pad_width for x in x.shape),
+        tuple(newshape),
         fill_value=value,
         dtype=x.dtype,
         device=_compat.device(x),
     )
-    padded[(slice(pad_width, -pad_width, None),) * x.ndim] = x
+    padded[tuple(slices)] = x
     return padded
 
 
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -596,7 +596,7 @@ def pad(
 
     # make pad_width a list of length-2 tuples of ints
     if isinstance(pad_width, int):
-        pad_width = [(pad_width, pad_width)] * x.ndim
+        pad_width = cast(list[tuple[int, int]], [(pad_width, pad_width)] * x.ndim)
 
     if isinstance(pad_width, tuple):
         pad_width = [pad_width] * x.ndim
@@ -812,7 +812,7 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
             raise ValueError(msg)
 
         if copy not in (True, False, None):
-            msg = f"copy must be True, False, or None; got {copy!r}"  # pyright: ignore[reportUnreachable]
+            msg = f"copy must be True, False, or None; got {copy!r}"
             raise ValueError(msg)
 
         if copy is None:
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -595,17 +595,17 @@ def pad(
     value = constant_values
 
     # make pad_width a list of length-2 tuples of ints
+    x_ndim = cast(int, x.ndim)
     if isinstance(pad_width, int):
-        pad_width = cast(list[tuple[int, int]], [(pad_width, pad_width)] * x.ndim)
-
+        pad_width = [(pad_width, pad_width)] * x_ndim
     if isinstance(pad_width, tuple):
-        pad_width = [pad_width] * x.ndim
+        pad_width = [pad_width] * x_ndim
 
     if xp is None:
         xp = array_namespace(x)
 
-    slices = []
-    newshape = []
+    slices: list[slice] = []
+    newshape: list[int] = []
     for ax, w_tpl in enumerate(pad_width):
         if len(w_tpl) != 2:
             msg = f"expect a 2-tuple (before, after), got {w_tpl}."
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -604,7 +604,8 @@ def pad(
     if xp is None:
         xp = array_namespace(x)
 
-    slices: list[slice] = []
+    # https://github.com/data-apis/array-api-extra/pull/82#discussion_r1905688819
+    slices: list[slice] = []  # type: ignore[no-any-explicit]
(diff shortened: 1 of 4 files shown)


src/array_api_extra/_funcs.py
```diff
@@ -568,8 +568,9 @@ def pad(
     ----------
     x : array
         Input array.
-    pad_width : int
+    pad_width : int or tuple of ints
         Pad the input array with this many elements from each side.
+        Ifa tuple, each element applies to the corresponding axis of `x`.
```
lucascolley, line 573:
```suggestion
        If a tuple, each element applies to the corresponding axis of `x`.
```

ev-br:
Testing https://github.com/scipy/scipy/pull/22122 and https://github.com/scipy/scipy/pull/22226 shows the need to bit the bullet and add `pad_width` being an int or a 2-tuple or a list of 2-tuples.
When round-tripping through numpy, it was all hidden out of sight; otherwise, need to implement this manually, as this PR does now.

src/array_api_extra/_funcs.py
```diff
@@ -555,7 +555,7 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
 
 def pad(
     x: Array,
-    pad_width: int,
+    pad_width: int | tuple | list,
```
lucascolley, line 558: I think CI is looking for something like

```suggestion
    pad_width: int | tuple[int, ...] | list[tuple[int, int], ...],
```

src/array_api_extra/_funcs.py
```diff
@@ -555,7 +555,7 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
 
 def pad(
     x: Array,
-    pad_width: int,
+    pad_width: int | tuple[int, int] | list[tuple[int, int], ...],
```
lucascolley, line 558: Second time lucky?

```suggestion
    pad_width: int | tuple[int, ...] | list[tuple[int, int]],
```

lucascolley:
The remaining mypy failures can be ignored with inline comments, I think.

ev-br:
I really don't understand what on earth it wants and how to silence it. Now taht you want all this on, I'd appreciate if you could silence it with whatever means you prefer as appropriate.
(in my projects, I'd just shut all this off and be done with it; you clearly prefer some other way)

lucascolley:
will do!

src/array_api_extra/_funcs.py
```diff
@@ -590,16 +594,43 @@ def pad(
...
+
+    if isinstance(pad_width, tuple):
+        pad_width = [pad_width] * x.ndim
+
     if xp is None:
         xp = array_namespace(x)
 
+    slices = []
+    newshape = []
+    for ax, w_tpl in enumerate(pad_width):
```
lucascolley, line 609: @detachhead Basedpyright seems to be missing this assignment on 599. With or without the explicit `cast` I've added here, we see:

```
/home/runner/work/array-api-extra/array-api-extra/src/array_api_extra/_funcs.py:609:32 - error: Argument of type "int | tuple[int, int] | list[tuple[int, int]]" cannot be assigned to parameter "iterable" of type "Iterable[_T@enumerate]" in function "__new__"
  Type "int | tuple[int, int] | list[tuple[int, int]]" is not assignable to type "Iterable[_T@enumerate]"
    "int" is incompatible with protocol "Iterable[_T@enumerate]"
      "__iter__" is not present (reportArgumentType)
```

https://github.com/data-apis/array-api-extra/actions/runs/12651016113/job/35250714798?pr=82

By line 609 `pad_width` must be a list. Is this a known issue?

lucascolley: adding the three casts in

```diff
diff --git a/src/array_api_extra/_funcs.py b/src/array_api_extra/_funcs.py
index d5ee7fe..8fa397b 100644
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -596,7 +596,7 @@ def pad(
 
     # make pad_width a list of length-2 tuples of ints
     if isinstance(pad_width, int):
-        pad_width = cast(list[tuple[int, int]], [(pad_width, pad_width)] * x.ndim)
+        pad_width = [(pad_width, pad_width)] * x.ndim
 
     if isinstance(pad_width, tuple):
         pad_width = [pad_width] * x.ndim
@@ -606,6 +606,7 @@ def pad(
 
     slices = []
     newshape = []
+    pad_width = cast(list[tuple[int, int]], pad_width)
     for ax, w_tpl in enumerate(pad_width):
         if len(w_tpl) != 2:
             msg = f"expect a 2-tuple (before, after), got {w_tpl}."
@@ -624,6 +625,8 @@ def pad(
         newshape.append(sh)
         slices.append(sl)
 
+    newshape = cast(list[int], newshape)
+    slices = cast(list[slice], slices)
     padded = xp.full(
         tuple(newshape),
         fill_value=value,
diff --git a/tests/test_funcs.py b/tests/test_funcs.py
index 4ae57ea..2fc9041 100644
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -426,7 +426,7 @@ class TestPad:
         assert padded.shape == (6, 7)
 
         with pytest.raises(ValueError, match="expect a 2-tuple"):
-            pad(a, [(1, 2, 3)])  # type: ignore[list-item]
+            pad(a, [(1, 2, 3)])  # type: ignore[list-item]  # pyright: ignore[reportArgumentType]
 
     def test_list_of_tuples_width(self):
         a = xp.reshape(xp.arange(12), (3, 4))
```

fixes the basedpyright errors, but now basedmypy complains about a redundant cast:

```
src/array_api_extra/_funcs.py:609:17: error: Redundant cast to "list[(int, int)]"  [redundant-cast]
        pad_width = cast(list[tuple[int, int]], pad_width)
```

lucascolley: @kotlinisland looks like there is something basedmypy is catching which basedpyright isn't? Any idea what?

DetachHead: it looks like pyright isn't narrowing it because `x` is an alias to `Any`. [here's a minified repro](https://basedpyright.com/?typeCheckingMode=all&reportAny=false&reportExplicitAny=false&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcAsAFDnkAmApsFAPoAUAbgIYA2ArjQFyxcEHGgG1UMADRQAdLIC6UAD5QOSAM4wxKGAuXipAD35E4ASl7koVzPXWoNbFAGMarTjynjzl67-bcaKABeKBF-HjlKMl8QGhYaTgZ4BFdwmlMoAGIVdU1xXQEhUX0ZeR8rcptMNXsYRxc3AKkYQWFvaN9rNODQgEYFACooA0rK2PjE5NT3dKzC4S1JUukC1Q1FgvEgA)

```python
from typing import Any


def _(value: tuple[int, ...] | list[int] | int, x: Any):
    if isinstance(value, int):
        value = [value]

    reveal_type(value) # list[int] | tuple[int, ...]
    
    if isinstance(value, tuple):
        value = [1] * x
    
    reveal_type(value) # tuple[int, ...] | list[int] | int
```

i think this behavior is more correct than what mypy is doing, because the `Any` could have an `__rmul__` that changes the type to something completely different. in fact i think this should probably widen `value` to `Any` instead because the type could be changed to anything.

KotlinIsland: here we can witness the mypy limitation in action:
https://mypy-play.net/?mypy=basedmypy-latest&python=3.12&gist=1fe93fa8969e127e80ae6f169613d7e6

lucascolley: the error is reproduced in CI: https://github.com/data-apis/array-api-extra/actions/runs/12654902636/job/35263897195?pr=82

lucascolley (approved):
thanks @ev-br, merging!

do let me know if you have a clue what is going on with the failure I can't reproduce in the playground @KotlinIsland !

src/array_api_extra/_funcs.py
lucascolley: thanks, that makes sense why I can't reproduce on the playground. With the following diff:

```
diff --git a/src/array_api_extra/_funcs.py b/src/array_api_extra/_funcs.py
index 2f4bc0f..0c9f3e3 100644
--- a/src/array_api_extra/_funcs.py
+++ b/src/array_api_extra/_funcs.py
@@ -605,7 +605,7 @@ def pad(
         xp = array_namespace(x)
 
     # https://github.com/data-apis/array-api-extra/pull/82#discussion_r1905688819
-    slices: list[slice] = []  # type: ignore[no-any-explicit]
+    slices: list[slice[int | None, int | None, None]] = []
     newshape: list[int] = []
     for ax, w_tpl in enumerate(pad_width):
         if len(w_tpl) != 2:
@@ -613,6 +613,7 @@ def pad(
             raise ValueError(msg)
 
         sh = x.shape[ax]
+        sl: slice[int | None, int | None, None]
         if w_tpl[0] == 0 and w_tpl[1] == 0:
             sl = slice(None, None, None)
         else:
```

I now see:

```
src/array_api_extra/_funcs.py:618:18: error: Incompatible types in assignment (expression has type "slice[int | Any, int | Any, int | Any]", variable has type "slice[int | None, int | None, None]") 
[assignment]
                sl = slice(None, None, None)
                     ^~~~~~~~~~~~~~~~~~~~~~~
src/array_api_extra/_funcs.py:623:18: error: Incompatible types in assignment (expression has type "slice[int | Any, int | Any, int | Any]", variable has type "slice[int | None, int | None, None]") 
[assignment]
                sl = slice(start, stop, None)
                     ^~~~~~~~~~~~~~~~~~~~~~~~
```

can you see where the `Any` is coming from? Perhaps something isn't working with `slice` being generic over `None`?

KotlinIsland:
```py
def __new__(
        cls, start: int | None, stop: int | None, step: int | None = None, /
    ) -> slice[int | MaybeNone, int | MaybeNone, int | MaybeNone]: ...
```
where `MaybeNone = Any`

and the reason `a: slice` contains explicit `Any` is that:

```py
_StartT = TypeVar("_StartT", covariant=True, default=Any)
_StopT = TypeVar("_StopT", covariant=True, default=Any)
_StepT = TypeVar("_StepT", covariant=True, default=Any)

class slice(Generic[_StartT, _StopT, _StepT]):
```

therefore:
```py
a: slice
# is the same as
a: slice[Any, Any, Any]
```

so this is an issue with typeshed: https://github.com/python/typeshed/issues/13376

PR #86, opened 2025-01-09
ENH: `lazy_apply`

crusaderky:
Wrapper around [jax.apply_pure_callback](https://jax.readthedocs.io/en/latest/_autosummary/jax.pure_callback.html) with added support for Dask.

CC @rgommers


.pre-commit-config.yaml
```diff
@@ -59,13 +59,13 @@ repos:
...
-        language: pygrep
-        entry: PyBind|Numpy|Cmake|CCache|Github|PyTest
-        exclude: .pre-commit-config.yaml
+  # - repo: local
+  #   hooks:
+  #     - id: disallow-caps
+  #       name: Disallow improper capitalization
+  #       language: pygrep
+  #       entry: PyBind|Numpy|Cmake|CCache|Github|PyTest
+  #       exclude: .pre-commit-config.yaml
```
crusaderky, line 68: Couldn't make it like `NumpyObject` and couldn't find a way to suppress the single error

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
+"""Public API Functions."""
+
+# https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
```
crusaderky, line 3: Out of scope: I think it's really unhealthy to unformally support Python 3.9, without any CI to back it.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+from __future__ import annotations
+
+from collections.abc import Callable, Hashable, Mapping, Sequence
+from functools import wraps
+from types import ModuleType
+from typing import TYPE_CHECKING, Any, cast
+
+if TYPE_CHECKING:
+    # https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
+    from typing import TypeAlias
```
crusaderky, line 13: Out of scope: I think it's really unhealthy to informally support Python 3.9, without any CI to back it.
For example, if I had accidentally put this line outside of TYPE_CHECKING it would have quietly broken it.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+        **Dask specific.**
+        Sequence of dicts, one per output, mapping index to function to be applied to
+        each chunk to determine the output size. The total must add up to the output
+        shape.
+        Default: on Dask, the size along each index cannot change.
+    new_axes : Sequence[Mapping[Hashable, int]], optional
+        **Dask specific.**
+        New indexes and their dimension lengths, one per output.
+        Default: on Dask, there can't be `output_indices` that don't appear in
+        `input_indices`.
```
crusaderky, line 93: All this is only needed for Dask.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+        `adjust_chunks` and `new_axes` depending on the function.
+
+        Read `dask.array.blockwise`:
+        - ``input_indices`` map to the even ``*args`` of `dask.array.blockwise`
+        - ``output_indices[0]`` maps to the ``out_ind`` parameter
+        - ``adjust_chunks[0]`` maps to the ``adjust_chunks`` parameter
+        - ``new_axes[0]`` maps to the ``new_axes`` parameter
+
+        ``core_indices`` is a safety measure to prevent incorrect results on
+        Dask along chunked axes. Consider this::
```
crusaderky, line 135: This design was informed from https://docs.xarray.dev/en/latest/generated/xarray.apply_ufunc.html

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+        This is treated as an implicit transfer.
+
+    PyTorch, CuPy:
+        These backends raise by default if you attempt to convert arrays on a GPU device
+        to NumPy.
+
+    Dask:
+        This function allows applying func to the chunks of dask arrays.
+        The dask graph won't be computed. As a special limitation, `func` must return
+        exactly one output.
```
crusaderky, line 122: I suspect that this limitation is straightforward to fix in Dask (at the cost of API duplication), but until then I suspect it will be a major roadblock for Dask support in scipy.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+
+        In order to allow Dask you need to specify at least
+        `input_indices`, `output_indices`, and `core_indices`, but you may also need
+        `adjust_chunks` and `new_axes` depending on the function.
+
+        Read `dask.array.blockwise`:
+        - ``input_indices`` map to the even ``*args`` of `dask.array.blockwise`
+        - ``output_indices[0]`` maps to the ``out_ind`` parameter
+        - ``adjust_chunks[0]`` maps to the ``adjust_chunks`` parameter
+        - ``new_axes[0]`` maps to the ``new_axes`` parameter
```
crusaderky, line 132: These are all lists for forward-compatibility to a yet-to-be-written da.blockwise variant that supports multiple outputs.

.pre-commit-config.yaml
lucascolley: it will want `NumPyObject`

src/array_api_extra/_apply.py
lucascolley: I spoke to some sklearn maintainers today and it sounded like dropping Python 3.9 soon is possible. I agree this isn't great, but at the same time, any issues would be caught in sklearn CI when trying to bump version.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+        tuple or list thereof.
+
+        It must be a pure function, i.e. without side effects such as disk output,
+        as depending on the backend it may be executed more than once.
+    *args : Array
+        One or more Array API compliant arrays.
+    shapes : Sequence[tuple[int, ...]], optional
+        Sequence of output shapes, one for each output of `func`.
+        If `func` returns a single (non-sequence) output, this must be a sequence
+        with a single element.
```
crusaderky, line 58: I tried overloading but it was a big headache when validating inputs. I found this approach much simpler.

lucascolley (commented):
I think I am missing some context here. Why are we wrapping arbitrary NumPy functions? Instead of, e.g., considering individual functions which we need one-by-one.

.pre-commit-config.yaml
crusaderky: bingo!

crusaderky:
> I think I am missing some context here. Why are we wrapping arbitrary NumPy functions? Instead of, e.g., considering individual functions which we need one-by-one.

There are many points in scipy that look like this:

```python
x = np.asarray(x)
y = np.asarray(y)
z = some_cython_kernel(x, y)
z = xp.asarray(z)
```

None of them will ever work with arrays on GPU devices, of course, and they'll either need a pure-array-api alternative or a dispatch to cupy.scipy etc.

None of them work with jitted, cpu-based JAX either, because `jax.jit` doesn't support abrupt materialization of the graph on `np.asarray` in the middle of jitting. Notably, this differs from `torch.compile`, which does exactly that.

With Dask, they technically work because there is no materialization guard but most times you would prefer if they crashed. In the best case they will be exceptionally slow to run once you reach production-sized data, as if there are multiple inputs chances are that large parts of the graph will be computed multiple times, and they trigger massive transfers between client and workers - which are very likely to kill off the client and the scheduler too (at least in the default configuration that sets direct_to_workers=False).

However, it is possible to make these pieces of Cython code work with thanks to this PR.
For JAX, this is straightforward - all you need to know is the output shape(s) and dtype(s).
For Dask, it is, well, the very opposite of straightforward. You *can* run arbitrary cython kernel in dask, on the workers, but with the very big caveat that any axis they reduce upon can't be chunked or you'll get incorrect behaviour, as explained in the docstring. Additionally, Dask needs to know, in order to function, how each axis of the input maps to each axis of the output, and if the size along each axis changes in any way, how that translates to chunk sizes.

There are two competing functions in Dask to achieve this, with different API:
- https://docs.dask.org/en/stable/generated/dask.array.map_blocks
- https://docs.dask.org/en/stable/generated/dask.array.blockwise.html

map_blocks is a variant of blockwise with simplified API, which can only work on broadcastable inputs.

This problem has already been dealt with by xarray, with https://docs.xarray.dev/en/latest/generated/xarray.apply_ufunc.html. Note that xarray API is more user-friendly thanks to each dimension being labelled at all times, so apply_ufunc can do fancy tricks like auto-transposing the inputs and pushing the dimensions that func doesn't know about to the left.

What I tried to implement here is equivalent to `xarray.apply_ufunc(..., dask="parallelized")`.
`dask="allowed"` makes no sense without an ulterior wrapper around the numpy-like API of dask.

lucascolley:
Okay, thanks. When you say

> any Array API compliant arrays

in the docstring, that isn't strictly true, right? It relies on `np.asarray` working on `xp` arrays and `xp.asarray` working on `np` arrays, if I have read the code correctly.

I understand the utility of this PR now.

> what's your vision for it?

I hadn't envisioned tackling this yet. In my mind getting Cython kernels working with dask/jax jit has been in the same "for later" basket as handling device transfers or writing new implementations to delegate to. But if the implementation works, makes sense to tackle it.

crusaderky:
> Okay, thanks. When you say
> 
> > any Array API compliant arrays
> 
> in the docstring, that isn't strictly true, right? It relies on `np.asarray` working on `xp` arrays and `xp.asarray` working on `np` arrays, if I have read the code correctly.

Correct. Nominally it will fail when densifying sparse arrays and moving data from GPU to CPU. A final user *can* however force their way through, if they want to, by deliberately suppressing transfer/densification guards for the time necessary to run the scipy function. Either that, or do an explicit device to cpu transfer / `to_dense()` call ahead of the function.

There is nothing however a jax or dask user can do today, short of completely getting out of the graph generation phase.

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+    xp: ModuleType | None = None,
+    input_indices: Sequence[Sequence[Hashable]] | None = None,
+    core_indices: Sequence[Hashable] | None = None,
+    output_indices: Sequence[Sequence[Hashable]] | None = None,
+    adjust_chunks: Sequence[dict[Hashable, Callable[[int], int]]] | None = None,
+    new_axes: Sequence[dict[Hashable, int]] | None = None,
+    **kwargs: Any,
+) -> tuple[Array, ...]:
+    """
+    Apply a function that operates on NumPY arrays to Array API compliant arrays.
```
rgommers, line 42: typo: NumPY -> NumPy

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,351 @@
...
+        This is treated as an implicit transfer.
+
+    PyTorch, CuPy
+        These backends raise by default if you attempt to convert arrays on a GPU device
+        to NumPy.
+
+    Sparse
+        By default, sparse prevents implicit densification through ``np.asarray`.
+        `This safety mechanism can be disabled
+        <https://sparse.pydata.org/en/stable/operations.html#package-configuration>`_.
```
rgommers, line 124: Fine to leave as is for now I'd say. Once `sparse` adds a better API for this (an env var doesn't work), it seems reasonable to add a `force=False` option to this function. There are various reasons why one may want to force an expensive conversion; that kind of thing should always be opt-in on a case-by-case basis.

rgommers (commented):
Thanks @crusaderky! 

> The enormous amount of extra complication needed to make it work with Dask makes me uncomfortable.

Yes indeed, it does. That looks like it's way too much. I don't think we would like to use all those extra keywords and Dask-specific functions in SciPy. If you'd drop those, does it make Dask completely non-working or is there a subset of functionality that would still work. I'd say that JAX shows that it can be straightforward, and a similar callback mechanism could be used for PyTorch/MLX/ndonnx as well - if that were to exist in those libraries.

rgommers:
Do you have a SciPy branch with this function being used @crusaderky? I'd be interested in playing with it.

crusaderky:
> If you'd drop those, does it make Dask completely non-working or is there a subset of functionality that would still work

I could make it work by rechunking all the inputs to a single chunk. In other words the whole calculation would need to fit in memory at once on a single worker.

crusaderky:
> I could make it work by rechunking all the inputs to a single chunk. In other words the whole calculation would need to fit in memory at once on a single worker.

@rgommers I rewrote it to do exactly this and now it's a lot cleaner. I'll keep my eyes open if I can see patterns in scipy we can leverage to improve Dask support (e.g. if there is there are frequent elementwise functions that could be trivially served by map_blocks)

src/array_api_extra/_apply.py
crusaderky: Shouldn't this be tackled by a more general design pattern?

```python
with disable_sensible_guards():
    y = scipy.somefunc(x)
```
applies to torch/cupy/jax arrays on GPU, as well as sparse arrays.

lucascolley:
@allcontributors, please add @crusaderky for bug

let me just try this once from this PR...

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,285 @@
+"""Public API Functions."""
+
+# https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
```
rgommers, line 3: nit: it'd be useful to have the actual reason here (`# Support Python 3.9`)

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,285 @@
...
+
+    Parameters
+    ----------
+    func : callable
+        The function to apply. It must accept one or more NumPy arrays or generics as
+        positional arguments and return either a single NumPy array or generic, or a
+        tuple or list thereof.
+
+        It must be a pure function, i.e. without side effects such as disk output,
+        as depending on the backend it may be executed more than once.
```
rgommers, line 76: This last line is a bit cryptic here, perhaps better to stop after "without side effects".

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,285 @@
...
+    elif is_jax_namespace(xp):
+        # If we're inside jax.jit, we can't eagerly convert
+        # the JAX tracer objects to numpy.
+        # Instead, we delay calling wrapped, which will receive
+        # as arguments and will return JAX eager arrays.
+
+        import jax  # type: ignore[import-not-found]  # pylint: disable=import-outside-toplevel,import-error  # pyright: ignore[reportMissingImports]
+
+        wrapped = _npfunc_wrapper(func, multi_output, xp)
+        out = cast(
```
rgommers, line 210: minor grumble: the level of typing/linting silencing and casts makes this code quite hard to read (nothing to do, just wish tools were better and there were fewer of them)

src/array_api_extra/_apply.py
```diff
@@ -0,0 +1,285 @@
...
+                **kwargs,
+            ),
+        )
+
+    else:
+        # Eager backends
+        wrapped = _npfunc_wrapper(func, multi_output, xp)
+        out = wrapped(*args, **kwargs)
+
+        # Output validation
```
rgommers, line 228: I hope this doesn't introduce too much overhead for NumPy arrays, there's a lot of checks here compared to a regular function call. May make sense to measure once you have a branch that uses this on some scipy functionality.

src/array_api_extra/_lib/_compat.pyi
```diff
@@ -19,8 +19,9 @@ def array_namespace(
 ) -> ArrayModule: ...
 def device(x: Array, /) -> Device: ...
 def is_cupy_namespace(x: object, /) -> bool: ...
+def is_dask_namespace(xp: ModuleType, /) -> bool: ...
 def is_jax_array(x: object, /) -> bool: ...
-def is_jax_namespace(x: object, /) -> bool: ...
+def is_jax_namespace(xp: ModuleType, /) -> bool: ...
```
rgommers, line 24: If you change two of them, best to change all `*__namespace` functions?

rgommers (commented):
I haven't tried to test this, but did go through it in a bit more detail now - overall looks good, a few comments. Looking forward to trying it out!

src/array_api_extra/_apply.py
crusaderky: I'm happy to remove this

src/array_api_extra/_apply.py
crusaderky: Yes, I fully agree. The overhead just to satisfy the linters is unjustified to me.

crusaderky:
I've reworked the design a bit.

- Renamed the function to `apply_lazy` and added a `as_numpy=False` optional parameter. This allows for
  - array API compliant eager functions to be wrapped by Dask and applied to Dask arrays with non-numpy `_.meta` , and
  - eager-only JAX operations (e.g. with output size that's not predictable) to be executed on lazy arrays, which is particularly beneficial for GPU 
- Added support for unknown output size in Dask and eager JAX. This allows Dask support for functions such as [scipy.cluster.leaders](https://github.com/scipy/scipy/blob/564da7926ac0900259bf62898f847b27252b6a49/scipy/cluster/hierarchy.py#L4160-L4171).

FYI, when I move `_lazywhere` from scipy, I intend to call it `apply_where`, which I think makes a good pair with `apply_lazy` as it conveys that they're both about applying a callback.

src/array_api_extra/_lib/_apply.py
```diff
@@ -0,0 +1,326 @@
...
+        If `as_numpy=False`, `func` will receive in input eager arrays of the meta
+        namespace, as defined by the `._meta` attribute of the input Dask arrays.
+        The outputs of `func` will be wrapped by the meta namespace, and then wrapped
+        again by Dask.
+
+    Raises
+    ------
+    jax.errors.TracerArrayConversionError
+        When `xp=jax.numpy`, `shape` is unknown (it contains None on one or more axes)
+        and this function was called inside `jax.jit`.
```
crusaderky, line 171: See comment below

crusaderky:
I think I want to have some proof that the whole thing works in practice before I finalize this PR.
See https://github.com/scipy/scipy/pull/22342

lucascolley:
(I added `backends` into the `lint` env recently, hence some mypy ignores have become unused)

src/array_api_extra/_lib/_apply.py
```diff
@@ -0,0 +1,326 @@
...
+        # as arguments and will return JAX eager arrays.
+
+        import jax  # type: ignore[import-not-found]  # pylint: disable=import-outside-toplevel,import-error  # pyright: ignore[reportMissingImports]
+
+        wrapped = _apply_lazy_wrapper(func, as_numpy, multi_output, xp)
+
+        if any(s is None for shape in shapes for s in shape):
+            # Unknown output shape. Won't work with jax.jit, but it
+            # can work with eager jax.
+            # Raises jax.errors.TracerArrayConversionError if we're inside jax.jit.
```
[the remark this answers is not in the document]
crusaderky: `scipy.cluster.leaders` is a function which in a nearby future will work in eager JAX but can't work in `jax.jit` short of a public API change, because its output arrays' shape is `xp.unique_values(input).shape`.

@pearu @vfdev-5 a while ago you asked offline how can we run inside jax.jit a function such as this. 
It will be possible for an end user to call do so, at the condition that they consume its output and revert to a known shape, for example:

```python
import array_api_extra as xpx
from scipy.cluster import leaders

def _eager(x):
    a, b = leaders(x)  # shapes = (None, ), (None, )
    xp = array_namespace(a, b)
    # silly example; probably won't make sense functionally
    return xp.max(a), xp.max(b)

@jax.jit
def f(x):
    return xpx.lazy_apply(
        _eager, x, shape=((), ()), dtype=(x.dtype, x.dtype))
    )
```

pyproject.toml
```diff
@@ -191,10 +192,6 @@ markers = [
 
 [tool.coverage]
 run.source = ["array_api_extra"]
-report.exclude_also = [
-    '\.\.\.',
-    'if TYPE_CHECKING:',
-]
```
crusaderky, line 197: This was causing coverage to skip the whole `_lazy.py` for some reason

src/array_api_extra/_lib/_lazy.py
```diff
@@ -0,0 +1,404 @@
...
+    elif is_jax_namespace(xp) and _is_jax_jit_enabled(xp):
+        # Delay calling func with jax.pure_callback, which will forward to func eager
+        # JAX arrays. Do not use jax.pure_callback when running outside of the JIT,
+        # as it does not support raising exceptions:
+        # https://github.com/jax-ml/jax/issues/26102
+        import jax
+
+        # Shield eager kwargs from being coerced into JAX arrays.
+        # jax.pure_callback calls jax.jit under the hood, but without the chance of
+        # passing static_argnames / static_argnums.
```
crusaderky, line 278: This was a very unpleasant discovery.

src/array_api_extra/_lib/_lazy.py
```diff
@@ -0,0 +1,404 @@
...
+        lazy_kwargs = {}
+        eager_kwargs = {}
+        for k, v in kwargs.items():
+            if _contains_jax_arrays(v):
+                lazy_kwargs[k] = v
+            else:
+                eager_kwargs[k] = v
```
crusaderky, line 285: The original design was *much* more simply assuming that there are no arrays in the kwargs.
But I later found out that there is severely obfuscated code in scipy that breaks this assumption.

src/array_api_extra/_lib/_lazy.py
```diff
@@ -0,0 +1,404 @@
...
+    del shape
+    del dtype
+    # End of shape and dtype parsing
+
+    # Backend-specific branches
+    if is_dask_namespace(xp):
+        import dask
+
+        metas = [arg._meta for arg in args if hasattr(arg, "_meta")]  # pylint: disable=protected-access
+        meta_xp = array_namespace(*metas)
```
crusaderky, line 247: FYI not being able to infer the meta for argument-less generator functions was the reason that made me ditch support for them. Alternatively I could have added a `meta_xp` parameter but I really disliked it as it would only be meaningful for dask.

src/array_api_extra/testing.py
```diff
@@ -256,7 +265,12 @@ def _allow_dask_compute(
...
         scheduler = CountingDaskScheduler(n, msg)
-        with dask.config.set({"scheduler": scheduler}):
-            return func(*args, **kwargs)
+        with dask.config.set({"scheduler": scheduler}):  # pyright: ignore[reportPrivateImportUsage]
+            out = func(*args, **kwargs)
+
+        # Block until the graph materializes and reraise exceptions. This allows
+        # `pytest.raises` and `pytest.warns` to work as expected. Note that this would
+        # not work on scheduler='distributed', as it would not block.
+        return dask.persist(out, scheduler="threads")[0]  # type: ignore[no-any-return,attr-defined,no-untyped-call,func-returns-value,index]  # pyright: ignore[reportPrivateImportUsage]
```
crusaderky, line 274: This is the important change in this module; the rest is cosmetic.

crusaderky:
This is ready for final review and merge.

src/array_api_extra/_lib/_lazy.py
```diff
@@ -0,0 +1,403 @@
...
+    Dask wrapping around other backends
+        If ``as_numpy=False``, `func` will receive in input eager arrays of the meta
+        namespace, as defined by the ``._meta`` attribute of the input Dask arrays.
+        The outputs of `func` will be wrapped by the meta namespace, and then wrapped
+        again by Dask.
+
+    Raises
+    ------
+    jax.errors.TracerArrayConversionError
+        When ``xp=jax.numpy``, `shape` is unknown (it contains None on one or more axes)
```
rgommers, line 183: I had to think about this for a while, and I suspect other readers may have to do the same. The question that "shape is unknown" brings up is how that happens. I guess it's a combo of data-dependent-shape in the input array (e.g., it's a return value from `unique_values`) and a lack of input validation on `shape`?

Did you consider checking this (`None in shape and is_iax_namespace(xp)`) and raising a more informative error than `TracerArrayConversionError` within a`lazy_apply`? I'm asking because I remember the `jax` exception not being very informative.

src/array_api_extra/_lib/_lazy.py
rgommers: Hmm, that's unfortunate. I guess the SciPy code was bad-but-not-fixable-nor-ignoreable?

rgommers (commented):
Thanks @crusaderky, and apologies for the long delay. I think this is good to go - a few questions left from me only. I think we should merge this, then update https://github.com/scipy/scipy/pull/22342 and see how we like that PR.

src/array_api_extra/_lib/_lazy.py
```diff
@@ -0,0 +1,403 @@
...
+        seen.add(id(x))  # pyright: ignore[reportUnknownArgumentType]
+
+        if isinstance(x, list) or type(x) is tuple:  # pylint: disable=unidiomatic-typecheck  # pyright: ignore[reportUnknownArgumentType]
+            return type(x)(recursion(i) for i in x)  # pyright: ignore[reportUnknownArgumentType]
+        if isinstance(x, tuple):  # namedtuple
+            return type(x)(*(recursion(i) for i in x))  # pyright: ignore[reportUnknownArgumentType]
+        # dict
+        return {k: recursion(v) for k, v in x.items()}
+
+    return recursion(x)
```
lucascolley, line 367: perhaps move this to `_helpers`?

lucascolley (commented):
I skimmed the diff quickly—happy to merge given Ralf's approval!

src/array_api_extra/_lib/_lazy.py
crusaderky: Fixable? Maybe, but it would require thorough understanding of everything that's going on in it and some sort of system that moves kwargs to args. I'll try giving it a second attempt.
Ignorable? definitely not; all the functions that use kwargs will fall over if you pass them an array otherwise.

src/array_api_extra/_lib/_lazy.py
crusaderky: > I guess it's a combo of data-dependent-shape in the input array (e.g., it's a return value from unique_values) and a lack of input validation on shape?

No, JAX arrays can't have None in the shape today. Functions like `unique_values` without the bespoke `size=` parameter, as well as `a[mask]`, flat out crash inside jax.jit.
This is about *this* function creating an unknown shape. I'll update the docstring to clarify.

Trivial example that reimplements `jnp.unique_values`:

```python
y = xpx.lazy_apply(np.unique_values, x, shape=(None, ), as_numpy=True)
```

> Did you consider checking this (None in shape and is_iax_namespace(xp)) and raising a more informative error than TracerArrayConversionError within alazy_apply?

Good point. Fixed.

src/array_api_extra/_lib/_lazy.py
crusaderky: Fixed :)

src/array_api_extra/_lib/_lazy.py
crusaderky: removed

crusaderky:
I've manged to avoid having arrays in the kwargs in scipy.
I've completely removed kwargs trawling here, and I added support for pure-python and None args to compensate.
It looks a lot cleaner and more robust now.

src/array_api_extra/_lib/_lazy.py
rgommers: Ah got it, thanks. Fix LGTM, so resolving this comment.

rgommers (approved):
Both the JAX and the Dask code paths look clean and understandable now - very nice!. No more comments, let's ship it I'd say!

rgommers:
Ah there is a merge conflict

lucascolley:
release inbound

rgommers:
Thanks Guido & Lucas!

PR #88, opened 2025-01-13
TST/BUG: run all tests on all backends; fix backend-specific bugs

crusaderky:
- Fix `create_diagonal` on JAX
- Fix `kron` on JAX, CuPy, and PyTorch
- Fix `pad` on JAX
- Fix `in1d` on PyTorch
- Run all tests vs. all backends
- Individually mark all outstanding failures
- New ~public~ private functions `xp_assert_equal` and `xp_assert_close`.
  Note that I have no plans to use these in scipy, as there are enough extra nuances
  there to warrant bespoke variants.


src/array_api_extra/_lib/_utils.py
```diff
@@ -54,7 +54,9 @@ def in1d(
     order = xp.argsort(ar, stable=True)
     reverse_order = xp.argsort(order, stable=True)
     sar = xp.take(ar, order, axis=0)
-    if sar.size >= 1:
+    ar_size = _compat.size(sar)
+    assert ar_size is not None, "xp.unique*() on lazy backends raises"
```
crusaderky, line 58: Caveat: this is missing https://github.com/data-apis/array-api-compat/pull/231

tests/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+from array_api_extra._lib._compat import device as get_device
+from array_api_extra._lib._typing import Device, ModuleType
+
+
+class Library(Enum):
+    """All array libraries explicitly tested by array-api-extra."""
+
+    ARRAY_API_STRICT = "array_api_strict"
+    NUMPY = "numpy"
+    NUMPY_READONLY = "numpy_readonly"
```
crusaderky, line 18: This runs numpy twice on all tests, which is less than ideal, but I could not find a cleaner generalized design.

tests/test_funcs.py
```diff
@@ -17,175 +15,184 @@
...
     sinc,
 )
-from array_api_extra._lib._typing import Array
+from array_api_extra._lib._compat import device as get_device
+from array_api_extra._lib._typing import Array, Device, ModuleType
+from array_api_extra.testing import xp_assert_close, xp_assert_equal
 
+from .conftest import Library
 
+# mypy: disable-error-code=no-untyped-usage
```
crusaderky, line 24: The return type of xp.anything is untyped, because xp is a ModuleType

tests/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+    CUPY = "cupy"
+    TORCH = "torch"
+    DASK_ARRAY = "dask.array"
+    SPARSE = "sparse"
```
crusaderky, line 22: For the record: there are ZERO tests in the whole test suite that run successfully on sparse, for a wide variety of reason.

src/array_api_extra/testing.py
lucascolley, line 1: could you detail if/how these functions deviate from the ones in `scipy._lib._array_api`? If we are going to expose these here, it would be good to be able to remove them from SciPy.

Also, consider https://github.com/data-apis/array-api-extra/issues/17#issuecomment-2463122633. But given we have since changed the scope of array-api-extra to include paths specific to known backends, I think they probably are in-scope now.

lucascolley (commented):
I looked over everything apart from the changes under `tests/` to start off with - seems like a good idea, thanks!

lucascolley:
Also, see https://github.com/data-apis/array-api-extra/pull/72 for my vision of how I would like to include delegation to library-specific implementations more broadly. The testing introduced here should be enough to push that PR over the line too 🎉

src/array_api_extra/testing.py
crusaderky: - no default_ctx context variable
- no ability to cherry-pick disabling shape, dtype, and/or namespace validation
- no special casing for 0d

All of these features _could_ be catered for here, but I feel they're all severely overdesigned for ensuring the health of array-api-extra.
My goal was to write something that serves well array-api-extra, and as a nice bonus allow other people to use it. 
I'm open to making the whole new module private if you'd prefer it that way.

lucascolley: let's keep it private for now and explain the situation in gh-17. If there are requests for them to be made public in the future, we should reconsider.

crusaderky:
Made the functions private.
codecov is (correctly) complaining about code paths specific to cupy and sparse.
Ready for final review and merge.

PR #92, opened 2025-01-15
TST: Run all tests on read-only numpy arrays

crusaderky:


tests/test_funcs.py
```diff
@@ -136,6 +136,7 @@ def test_device(self, xp: ModuleType, device: Device):
         x = xp.asarray([1, 2, 3], device=device)
         assert get_device(cov(x)) == device
 
+    @pytest.mark.skip_xp_backend(Backend.NUMPY_READONLY)
```
crusaderky, line 139: Notably, read-only numpy disallows `__iadd__` etc., whereas jax and sparse simply don't define these methods, which causes them to fall back to `a = a.__add__(b)`

tests/test_utils.py
```diff
@@ -16,10 +15,10 @@ class TestIn1D:
     @pytest.mark.skip_xp_backend(Backend.DASK_ARRAY, reason="no argsort")
     @pytest.mark.skip_xp_backend(Backend.SPARSE, reason="no unique_inverse, no device")
     # cover both code paths
-    @pytest.mark.parametrize("x2", [np.arange(9), np.arange(15)])
-    def test_no_invert_assume_unique(self, xp: ModuleType, x2: Array):
+    @pytest.mark.parametrize("n", [9, 15])
+    def test_no_invert_assume_unique(self, xp: ModuleType, n: int):
         x1 = xp.asarray([3, 8, 20])
-        x2 = xp.asarray(x2)
+        x2 = xp.arange(n)
```
crusaderky, line 21: `NumPyReadOnly.asarray(x2, copy=None)` was tainting the globally defined x2, ruining it for the following iterations of the test.

tests/conftest.py
```diff
@@ -34,6 +42,56 @@ def library(request: pytest.FixtureRequest) -> Backend:  # numpydoc ignore=PR01,
...
+class NumPyReadOnly:
+    """
+    Variant of array_api_compat.numpy producing read-only arrays.
+
+    Note that this is not a full read-only Array API library. Notably,
+    array_namespace(x) returns array_api_compat.numpy, and as a consequence array
+    creation functions invoked internally by the tested functions will return
+    writeable arrays, as long as you don't explicitly pass xp=xp.
+    For this reason, tests that do pass xp=xp may misbehave and should be skipped
+    for NUMPY_READONLY.
```
lucascolley, line 54: not blocking at all, but would it be possible to define `__array_namespace__` here to circumvent this?

lucascolley (approved):
seems good, thanks!

tests/conftest.py
lucascolley: I suppose not, since array-api-compat checks `is_numpy_array` before `hasattr(x, '__array_namespace__'), and it has to do that since NumPy defines `__array_namespace__` while still requiring wrapping.

crusaderky: We *want* tested functions to use xp=np. Otherwise as soon as they call things like `mask = xp.ones(...); mask &= ...` if will fall apart. I've reworked the docstring to make it clear.

lucascolley: that's clear now, thanks!

PR #101, opened 2025-01-19
MAINT: various refactoring

lucascolley:


crusaderky:
This PR makes it impossible to have a line with ``import jax`` and run the linters on it on Windows, where there are no JAX builds available yet.

crusaderky:
It also makes running `pixi run -e lint lint` a *very* long affair after every time you need to wipe your mypy cache. Which, thanks to based-mypy's bugs, I find myself doing on a regular basis.

lucascolley:
Ah, very good point, sorry about that.

I think we should be able to fallback to JAX from PyPI on windows.

crusaderky:
> If the times are too long then maybe it is better to revert to using import ignores, though.

Yes please.

lucascolley:
For some context, my motivation was getting the lint task to work in the `dev` env. If we can isolate the lint feature to just the lint env, that may be unnecessary.

crusaderky:
Ah yes, I've been fruitlessly trying to skip the mandatory `-e lint` too.

PR #104, opened 2025-01-20
DEV: `pixi run lint` improvements

crusaderky:
- Make `pixi run lint` much faster and allow running it on Windows (partially reverts #101)
- ~No need to explicitly specify lint environment anymore~
- Rename some environments for the sake of UX friendliness when a user runs `pixi run tests`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,9 @@ jobs:
           environments: lint
       - name: Run Pylint, Mypy & Pyright
         run: |
-          pixi run pylint
-          pixi run mypy
-          pixi run pyright
+          pixi run -e lint pylint
+          pixi run -e lint mypy
+          pixi run -e lint pyright
 
   checks:
     name: Check ${{ matrix.environment }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,9 @@ jobs:
           environments: lint
       - name: Run Pylint, Mypy & Pyright
         run: |
-          pixi run -e lint pylint
-          pixi run -e lint mypy
-          pixi run -e lint pyright
+          pixi run pylint
+          pixi run mypy
+          pixi run pyright
 
   checks:
     name: Check ${{ matrix.environment }}
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        environment: [ci-py310, ci-py313, ci-backends]
+        environment: [tests-py310, tests-py313, tests-backends]
         runs-on: [ubuntu-latest]
 
     steps:
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -136,4 +136,8 @@ pixi shell -e lint
 
 If you run on a host with CUDA hardware, you can enable extra tests:
 
-pixi shell -e dev-cuda pixi shell -e tests-cuda pixi run -e tests-cuda tests
+```
+pixi shell -e dev-cuda
+pixi shell -e tests-cuda
+pixi run -e tests-cuda tests
+```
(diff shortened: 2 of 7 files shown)


lucascolley:
thanks for fixing this up!!

Could you explain why we want `lint-env` in the `dev` environment? IIUC, even after `pixi shell -e dev`, `pixi run lint` should temporarily fire up the `lint` env to run the task?

lucascolley:
we should also update the contributor docs, but happy to leave that to a follow up if you don't want to do it here

crusaderky:
Ah. It doesn't work at all anymore.
```
(array-api-extra:dev) crusaderky@melody:~/github/array-api-extra$ pixi run lint
lint: command not found

Available tasks:
        clean-vendor-compat
        clean-vendor-extra
        copy-vendor-compat
        copy-vendor-extra
        coverage
        docs
        ipython
        open-coverage
        open-docs
        tests
        tests-ci
        tests-cov
        tests-vendor
```

Would it be reasonable to run `pixi run lint` only outside of the dev shell?
If not, I can revert that bit.

lucascolley:
Hmmm. Could we get rid of the `dev` environment entirely and have each of the test tasks only be available in one environment?

I'm really not sure what would be most idiomatic for pixi here. @ruben-arts is it idiomatic to have a task like `tests` which can be run from multiple environments, or is it best to specify different tasks? We find ourselves wanting to `pixi shell -e dev` so that we can `pixi run tests` without having to select the regular development environment every time. But we would also like to `pixi run lint` without having to specify the (orthogonal) environment each time too.

I know that there are ways to do this by bootstrapping tasks to call tasks in other environments, but I think I'd rather avoid that 😅.

crusaderky:
I've re-enabled `pixi run lint` from the `dev` shell. We can easily change it the other way around if you prefer.

lucascolley (approved):
thanks @crusaderky, this is a solid improvement now. Would be nice to follow-up if we can find a way to avoid having to specify the environment every time for one of the tasks, but that can wait I guess :)

ruben-arts:
Hi @lucascolley

What I like most is to define tasks only in the features, and only make myself write the `-e` when I have to run something special or in automated scripts like CI.

Sometimes this means that you have to make the `default` environment the `dev` environment. Which is what I often do.

e.g. what you could do is something like this:
```toml
[tasks]
default = "echo default"

[feature.test.tasks]
test = "echo test"

[feature.lint.tasks]
lint = "echo lint"

[environments]
default = {features = ["test", "lint"], solve-group = "default"}
test = {features = ["test"], solve-group = "default"}
lint = { features = ["lint"], solve-group = "default"}
```
Then in CI run this to minimize the environments
- `pixi run -e test test`
- `pixi run -e lint lint`

But locally you will have one big know it all environment that can run anything, this also helps with IDE support as you don't require different environments for different code paths.

Let me know if this would make sense to you and if you found issues with this approach I should be aware of!

lucascolley:
thanks for the quick response!!

> But locally you will have one big know it all environment that can run anything, this also helps with IDE support as you don't require different environments for different code paths.
> 
> Let me know if this would make sense to you and if you found issues with this approach I should be aware of!

Since `mypy`'s behaviour differs based on which array backend libraries we have installed, we would ideally include all of those backends in the lint environment, otherwise we have to add `ignore` comments which need to be allowed to be unused depending on which env we run the lint task from.

I think the main concern is `pixi run lint` being too slow from the default environment, mainly due to `mypy` itself already being quite slow, and getting worse the more libraries from which we import are available. For now we would be happy for the `lint` task to not be in an env with those optional dependencies, but then we are unable to run the task from the "run anything" environment without specifying `-e`...

PR #105, opened 2025-01-20
TST: better pytest short summary info

crusaderky:
Change the "short summary info" at the end of a pytest run to be more informative re. which backends have triggered skips


lucascolley (approved):
nice, thanks!

PR #113, opened 2025-01-20
ENH: new function `isclose`

crusaderky:


tests/test_funcs.py
```diff
@@ -23,7 +25,7 @@
 from array_api_extra._lib._utils._typing import Array, Device
 
 # some xp backends are untyped
-# mypy: disable-error-code=no-untyped-usage
+# mypy: disable-error-code=no-untyped-def
```
crusaderky, line 28: This is to fix vscode's mypy plugin, which otherwise highlights the whole module in red.

lucascolley (commented):
Please could you read through the discussion at https://github.com/data-apis/array-api/issues/170 and let me know how this PR stands in relation to that?

crusaderky:
> Please could you read through the discussion at [data-apis/array-api#170](https://github.com/data-apis/array-api/issues/170) and let me know how this PR stands in relation to that?

This is 1:1 what numpy, cupy, tensorflow, pytorch, jax and dask already implement.

- it uses the controversial, but backwards compatible, numpy algorithm instead of the `math.isclose` one.
- it uses the libraries' default `atol` and `rtol` (which are problematic for small numbers) instead of the better ones from `math.isclose`, again for the sake of backwards compatibility.
- it doesn't tweak tolerances for float32, like `assert_allclose` does, for the same reason.
- it applies the same default tolerances to integers. I found myself in several occasions to find this useful.
- it treats bools as integers, like numpy does. This is IMHO unnecessary but I had to special-case it either way (either in `_func` to respect numpy's behaviour or in `_delegation` to deviate from it even when delegation is normally possible). I think it's enough of an edge case that either decision is inconsequential.
- there is discussion in the RFC about the usefulness of `equal_nan=False`, with consensus veering on it being frequently very useful. The RFC fails to highlight that you need scipy's `lazywhere` in order to suppress the warnings. Again, this PR implements what numpy does; however for array-api-strict (and possibly other non-special-cased libraries that wrap around numpy) you temporarily still have spurious warnings. I will fix this in the PR that adds `lazywhere`.
- `inf` behaviour is the same as all libraries as well as math.isclose and in line with the recommendations of the RFC (`inf` is only close to itself, as is `-inf`).

The RFC is right to point out though that `allclose` is trivial and overkill to implement. I've removed it from this PR.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -305,6 +305,41 @@ def expand_dims(
...
+        # FIXME: use scipy's lazywhere to suppress warnings on inf
+        out = xp.abs(a - b) <= (atol + rtol * xp.abs(b))
+        out = xp.where(xp.isinf(a) & xp.isinf(b), xp.sign(a) == xp.sign(b), out)
+        if equal_nan:
+            out = xp.where(xp.isnan(a) & xp.isnan(b), xp.asarray(True), out)
+        return out
+
+    if xp.isdtype(a.dtype, "bool") or xp.isdtype(b.dtype, "bool"):
+        if atol >= 1 or rtol >= 1:
+            return xp.ones_like(a == b)
```
crusaderky, line 332: On eager backends, this is less performant than
```python
return xp.ones(xp.broadcast_arrays(a, b)[0], dtype=bool, device=a.device)
```
but it supports backends with NaN shapes like Dask.
jax.jit elides it away.

src/array_api_extra/_lib/_testing.py
```diff
@@ -40,8 +42,16 @@ def _check_ns_shape_dtype(
...
+    actual_shape = actual.shape
+    desired_shape = desired.shape
+    if is_dask_namespace(desired_xp):
+        if any(math.isnan(i) for i in actual_shape):
+            actual_shape = actual.compute().shape
+        if any(math.isnan(i) for i in desired_shape):
+            desired_shape = desired.compute().shape
+
+    msg = f"shapes do not match: {actual_shape} != f{desired_shape}"
+    assert actual_shape == desired_shape, msg
```
crusaderky, line 54: This will have to be replicated on the scipy PR for dask support

rgommers:
> This is 1:1 what numpy, cupy, tensorflow, pytorch, jax and dask already implement.

That seems like the right choice to me for `array-api-extra`.

src/array_api_extra/_delegation.py
```diff
@@ -30,6 +30,95 @@ def _delegate(xp: ModuleType, *backends: Backend) -> bool:
...
+    with default settings. Be sure to select atol for the use case at hand, especially
+    for defining the threshold below which a non-zero value in `a` will be considered
+    "close" to a very small or zero value in `b`.
+
+    The comparison of `a` and `b` uses standard broadcasting, which means that `a` and
+    `b` need not have the same shape in order for `isclose(a, b)` to evaluate to
+    `True`.
+
+    `isclose` is not defined for non-numeric data types. `bool` is considered a numeric
+    data-type for this purpose.
```
lucascolley, line 105:
```suggestion
    `isclose` is not defined for non-numeric data types. ``bool`` is considered a numeric
    data-type for this purpose.
```

lucascolley:
happy to merge once unresolved comments are addressed

crusaderky:
> happy to merge once unresolved comments are addressed

all done

PR #114, opened 2025-01-20
ENH: `pad`: `pad_width` can be any sequence

crusaderky:
This is a blocker to porting https://github.com/scipy/scipy/pull/22308 to xpx, because jax.jit doesn't accept unhashable static arguments.

For clarity: this is just a quirk of https://github.com/scipy/scipy/pull/22308. You *can* pass a list parameter to `xpx.pad` inside jax.jit; you simply can't pass it if you are calling `pad = jax.jit(xpx.pad, static_argnames="pad_width")`, which is not something final users will typically do.


lucascolley (approved):
thanks!

PR #121, opened 2025-01-22
ENH: `at` support for bool mask in Dask and JAX

crusaderky:
This special case treatment should be incorporated inside jax itself.
In the meantime, for the sake of expedience, I'm implementing it here.


src/array_api_extra/_lib/_at.py
```diff
@@ -141,6 +146,25 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
...
+        >>> x[mask] = f(x[mask])
+
+    Can't be replaced by `at`, as it won't work on Dask and JAX inside jax.jit::
+
+        >>> mask = m(x)
+        >>> x = xpx.at(x, mask).set(f(x[mask])  # Crash on Dask and jax.jit
+
+    You should instead use::
+
+        >>> x = xp.where(m(x), f(x), x)
```
crusaderky, line 166: TODO suggest to use lazywhere

crusaderky:
@lucascolley ready for review!

src/array_api_extra/_lib/_at.py
```diff
@@ -141,6 +146,25 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
     not explicitly covered by ``array-api-compat``, are not supported by update
     methods.
 
+    Boolean masks are supported on Dask and jitted JAX arrays exclusively
+    when `idx` has the same shape as `x` and `y` is 0-dimensional.
+    Note that this is support is not available in JAX's native
```
lucascolley, line 151:
```suggestion
    Note that this support is not available in JAX's native
```

tests/test_at.py
```diff
@@ -192,6 +191,62 @@ def test_iops_incompatible_dtype(op: _AtOp, copy: bool):
...
+    """
+    x = xp.asarray([1, 2])
+    idx = xp.asarray([True, False])
+    if library in (Backend.DASK, Backend.JAX):
+        z = at_op(x, idx, op, 1.1)
+        assert z.dtype == x.dtype
+
+    else:
+        with pytest.raises(Exception, match=r"cast|promote|dtype"):
+            at_op(x, idx, op, 1.1)
```
lucascolley, line 229: Higher level question - what is the point of introducing behaviour which works for Dask and JAX but not other backends?

crusaderky: I'm not a fan of it either. Can you think of what the test should be to make them fail like the others?

lucascolley: Maybe the docstring of this test could be worded differently. Saying "must retain x.dtype" seems to imply that it must succeed without error for all backends. If we instead phrase it like you have above saying that it "replicates the backend's behaviour", maybe nothing else needs to change.

crusaderky: I've rewritten the test

tests/test_at.py
```diff
@@ -176,22 +173,75 @@ def test_alternate_index_syntax():
...
+
     >>> a = np.asarray([1, 2, 3])
     >>> a / 1.5
     array([0.        , 0.66666667, 1.33333333])
     >>> a /= 1.5
     UFuncTypeError: Cannot cast ufunc 'divide' output from dtype('float64')
     to dtype('int64') with casting rule 'same_kind'
+
+    See Also
+    --------
```
lucascolley, line 200:
```suggestion
```

lucascolley (approved):
thanks, LGTM!

PR #124, opened 2025-01-24
WIP ENH: `setdiff1d` for Dask and jax.jit

### LINKED ISSUE #116: BUG: `setdiff1d` for jax.jit and dask NaN-shaped arrays
Follow-up to https://github.com/data-apis/array-api-extra/pull/115

- `setdiff1d` fails when executed inside jax.jit. This is because a call to `xp.unique*` without an explicit size= parameter.
- By reading the code I'm fairly sure that the function fails if x1 and/or x2 have Dask backend and NaN in their shapes. Lacks a unit test.

crusaderky:
Closes #116
Needs more thorough unit tests + performance benchmarks.

This function's output is of unknown shape, so with the previous API it will never work in jax.jit.

There are a few options:
1. Stipulate that the function will never work inside `jax.jit` and you need to hack your way around it with https://github.com/data-apis/array-api-extra/pull/86#discussion_r1925702055.
I'm not a fan of this because UX is very painful as it forces the user to think in graphs.
2. Add optional parameter `fill_value`. iff running inside the jax.jit, quietly return a longer array padded with it.
I'm not happy about this because it causes jax.jit to quietly diverge from other backends and users will spend a lot of time debugging.
3. Add optional parameters `size` and `fill_value`. `size` becomes mandatory when running inside jax.jit. This is the same design as `jax.numpy.unique_values`.
This also allows having a known-shape output in Dask. However, implementing it for Dask is fairly complicated.
4. Add optional parameters `size` and `fill_value`. `size` is mandatory when running inside jax.jit and disregarded otherwise. Again, this will cause bugs in the user code that only appear in jax.jit, but at least it demands an initial explicit user intervention. This is the simplest to implement; unsure on the UX. It also has the advantage of not sacrificing performance on other backends. If in the future jax.jit will support arrays of unknown size, it becomes easy to deprecate it as we said that the output size requested by the user may be disregarded anyway.

My current favourite is (4).
@rgommers you previously said, talking about functions with the same problem in scipy, that you prefer (1) to (3) because of not being able to retract the API in the future. What's your opinion on (4)?

CC @lucascolley

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -9,7 +9,6 @@ environments:
       linux-64:
       - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-2_kmp_llvm.tar.bz2
-      - conda: https://prefix.dev/conda-forge/noarch/array-api-compat-1.10.0-pyhd8ed1ab_0.conda
       - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-h4bc722e_7.conda
       - conda: https://prefix.dev/conda-forge/linux-64/ca-certificates-2024.12.14-hbcca054_0.conda
       - conda: https://prefix.dev/conda-forge/linux-64/ld_impl_linux-64-2.43-h712a8e2_2.conda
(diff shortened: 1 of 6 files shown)


crusaderky:
@rgommers confirmed offline his preference for delaying indefinitely. His reasoning is that _at some point_ JAX should support unknown shapes and the issue of niche functions that can't work until then should be marginal enough not to warrant urgent attention.

lucascolley:
> issue of niche functions that can't work until then should be marginal enough

Worth noting that this was hit for Dask in the cross-linked sklearn PR

lucascolley:
> @rgommers confirmed offline his preference for delaying indefinitely. His reasoning is that _at some point_ JAX should support unknown shapes and the issue of niche functions that can't work until then should be marginal enough not to warrant urgent attention.

it would be nice to have an upstream issue at which to point, however 'at some point' it is

PR #129, opened 2025-01-29
BUG: `setdiff1d` vs. non-1d arrays

crusaderky:
Fix bug in setdiff1d when x2 is not 1-dimensional and assume_unique=True


lucascolley (approved):
thanks!

PR #130, opened 2025-01-29
BUG: `isclose` integer overflow

crusaderky:
Fix crash in `isclose`, which with the default rtol happens when b.dtype is (u)int8 or (u)int16

### DIFF
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -335,7 +335,12 @@ def isclose(
     atol = int(atol)
     if rtol == 0:
         return xp.abs(a - b) <= atol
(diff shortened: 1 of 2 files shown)


lucascolley (commented):
should this be documented (that the `rtol` is ignored in this case)? Perhaps we can also emit a warning if the user supplies an `rtol` that would be ignored?

crusaderky:
> should this be documented (that the `rtol` is ignored in this case)? Perhaps we can also emit a warning if the user supplies an `rtol` that would be ignored?

It's not ignored. 65535 * 1e-5 < 1, so you can never have a False that flips to True because of it.

lucascolley (approved):
makes sense, thanks!

PR #131, opened 2025-01-29
ENH: Array API 2024.12 binary ops vs. Python scalars

crusaderky:
Implement 2024.12 rules for `kron`, `isclose`, and `setdiff1d` for when a binary operation runs on an Array API object and a python scalar.

https://data-apis.org/array-api/draft/API_specification/type_promotion.html#mixing-arrays-with-python-scalars

### DIFF
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -359,7 +359,7 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType | None = None) -> Array:
     Parameters
     ----------
     a, b : Array | int | float | complex
-        Input arrays or scalars. At least one must be an Array API object.
+        Input arrays or scalars. At least one must be an array.
     xp : array_namespace, optional
         The standard-compatible namespace for `a` and `b`. Default: infer.
 
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -12,6 +12,7 @@ from typing import cast
 from ._at import at
 from ._utils import _compat, _helpers
 from ._utils._compat import array_namespace, is_jax_array
+from ._utils._helpers import asarrays
 from ._utils._typing import Array
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


src/array_api_extra/_lib/_funcs.py
```diff
@@ -414,10 +416,10 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType | None = None) -> Array:
     """
     if xp is None:
         xp = array_namespace(a, b)
+    a, b = asarrays(a, b, xp=xp)
 
-    b = xp.asarray(b)
     singletons = (1,) * (b.ndim - a.ndim)
-    a = xp.broadcast_to(xp.asarray(a), singletons + a.shape)
```
crusaderky, line 420: btw this was pointless as we were calling b.ndim one line above

tests/test_funcs.py
```diff
@@ -467,6 +477,18 @@ def test_kron_shape(
...
+        a = 1
+        # Test no dtype promotion to xp.asarray(a); use b.dtype
+        b = xp.asarray([[1, 2], [3, 4]], dtype=xp.int16)
+        xp_assert_equal(kron(a, b), b)
+        xp_assert_equal(kron(b, a), b)
+        xp_assert_equal(kron(1, 1, xp=xp), xp.asarray(1))
+
+    def test_all_python_scalars(self):
+        with pytest.raises(TypeError, match="Unrecognized"):
+            kron(1, 1)
```
crusaderky, line 490: This does NOT fail if xp=jax.numpy, because xp_lazy_function converts everything to jax.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -356,8 +358,8 @@ def kron(a: Array, b: Array, /, *, xp: ModuleType | None = None) -> Array:
 
     Parameters
     ----------
-    a, b : array
-        Input arrays.
+    a, b : Array | int | float | complex
+        Input arrays or scalars. At least one must be an Array API object.
```
lucascolley, line 362:
```suggestion
        Input arrays or scalars. At least one must be an array.
```

PR #132, opened 2025-02-05
TST: `xfail_xp_backend`

crusaderky:
- Use `skip_xp_backend` when the library is Array API compliant, but with rightful limitations that make the functionality fail
- Use `xfail_xp_backend` when the failure is caused by non-compliance with the Array API. The test runs anyway; `pixi run tests` will exit 1 if there are any `XPASS` tests.


lucascolley (approved):
LGTM, thanks!

PR #133, opened 2025-02-05
ENH: `broadcast_shapes`

crusaderky:
`lazy_apply` benefits from this.

### DIFF
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -72,6 +72,8 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
     return x
 
 
+# `float` in signature to accept `math.nan` for Dask.
+# `int`s are still accepted as `float` is a superclass of `int` in typing
 def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...]:
     """
     Compute the shape of the broadcasted arrays.
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -17,6 +17,7 @@ from ._utils._typing import Array
 
 __all__ = [
     "atleast_nd",
+    "broadcast_shapes",
     "cov",
     "create_diagonal",
     "expand_dims",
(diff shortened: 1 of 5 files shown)


src/array_api_extra/_lib/_funcs.py
```diff
@@ -71,6 +72,67 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
     return x
 
 
+def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...]:
```
lucascolley, line 75:
```suggestion
def broadcast_shapes(*shapes: tuple[int | None, ...]) -> tuple[int | None, ...]:
```

tests/test_funcs.py
```diff
@@ -113,6 +115,62 @@ def test_xp(self, xp: ModuleType):
...
+        with pytest.raises(ValueError, match=match):
+            _ = broadcast_shapes(*args)
+
+    @pytest.mark.parametrize(
+        "args",
+        [
+            ((None,), (None,)),
+            ((math.nan,), (None,)),
+            ((1, None, 2, 4), (2, 3, None, 1), (2, None, None, 4)),
+            ((1, math.nan, 2), (4, 2, 3, math.nan), (4, 2, None, None)),
```
lucascolley, line 165: Not that there is a library for which you would hit this in practice, but we could add a test where the input has _both_ `None` and `math.nan` dimensions? Feel free to ignore if that seems pointless.

lucascolley (approved):
thanks @crusaderky, LGTM!

src/array_api_extra/_lib/_funcs.py
crusaderky: Not a typo. It is to allow for Dask's NaNs. Alas `Literal[math.nan]` is not a valid type.

tests/test_funcs.py
crusaderky: Good point. One may use the value returned by a previous call to `broadcast_shapes` with a shape of a dask array. Added it.

src/array_api_extra/_lib/_funcs.py
lucascolley: Do we want `float | int | None` then?

crusaderky: in typing (but not at runtime), float is a superclass of int https://peps.python.org/pep-0484/#the-numeric-tower

lucascolley:
```suggestion
# `float` in signature to accept `math.nan` for Dask.
# `int`s are still accepted as `float` is a superclass of `int` in typing
def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...]:
```

PR #135, opened 2025-02-06
BUG: `at` should not force overwrite in Dask when copy=None

### LINKED ISSUE #134: `at.set` behavior for Dask arrays?
Re: https://github.com/scipy/scipy/pull/22469#discussion_r1943450104

I encountered some code involving Dask and `at.set` that works as expected in 0.6.0 but not in the more recent commit used by SciPy (c23ac01).

```python3
import numpy as np
import dask.array as xp
from scipy._lib._array_api import array_namespace

# import array_api_extra as xpx  # this works
import scipy._lib.array_api_extra as xpx  # this doesn't

x = xp.asarray([1., 2., 3.])
y = xp.asarray([1., 2.5, 3.])
xp = array_namespace(x, y)

r = xp.vecdot(x, y)
mask = xp.asarray(False)
y = xp.asarray(xp.nan)

# r[mask] = xp.nan  # This works
r = xpx.at(r, mask).set(y)  # This doesn't

np.asarray(r)
# TypeError: 'numpy.float64' object does not support item assignment
```

But this brings up a question - why does `array_api_extra.at.set` get involved with Dask here anyway?  The documentation of Dask Array says:
image

The last line is what's relevant. Doing the regular indexing assignment (when `y` is a scalar, at least) seems to be something Dask can do, so why replace that with a `where`?

crusaderky:
Closes #134

- When x is a dask array,
- and idx is a bool mask of the same shape as x,
- and y is a scalar,
- and copy=None,

don't perform the unnecessary extra step of writing back to `x` (which encroaches on https://github.com/dask/dask/issues/11722).

This PR works around a bug of Dask 2025.1.0. Starting from Dask 2025.2.0, this becomes a nice-to-have.

### DIFF
--- a/tests/test_at.py
+++ b/tests/test_at.py
@@ -3,7 +3,7 @@ import pickle
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
 from types import ModuleType
-from typing import Any, cast
+from typing import cast
 
 import numpy as np
 import pytest
@@ -23,12 +23,13 @@ pytestmark = [
 ]
 
 
-def at_op(  # type: ignore[no-any-explicit]
+def at_op(
     x: Array,
     idx: Index,
     op: _AtOp,
     y: Array | object,
-    **kwargs: Any,  # Test the default copy=None
+    copy: bool | None = None,
+    xp: ModuleType | None = None,
 ) -> Array:
     """
     Wrapper around at(x, idx).op(y, copy=copy, xp=xp).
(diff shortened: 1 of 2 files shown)


tests/test_at.py
```diff
@@ -119,29 +121,65 @@ def assert_copy(array: Array, copy: bool | None) -> Generator[None, None, None]:
...
+    x = xp.asarray([1.0, 10.0, 20.0])
+    expect_copy = not is_writeable_array(x)
+    meth = cast(Callable[..., Array], getattr(at(x)[:2], op.value))  # type: ignore[no-any-explicit]
+    with assert_copy(x, None, expect_copy):
+        _ = meth(2.0)
+
+    x = xp.asarray([1.0, 10.0, 20.0])
+    # Dask's default copy value is True for bool masks,
+    # even if the arrays are writeable.
+    expect_copy = not is_writeable_array(x) or library is Backend.DASK
```
lucascolley, line 176: is this likely to remain the case indefinitely?

crusaderky: I don't see why we should remove it. Even without Dask <=2025.1.0's bug, this PR removes an unnecessary operation. `copy=None` means "do whatever is best for the library and the situation...

PR #137, opened 2025-02-07
ENH: `create_diagonal`: support broadcasting

### LINKED ISSUE #136: Make create_diagonal support broadcasting
Thus, the docstring would be something like:
```
    """A vectorized version of diagonal.

    Args:
        m: Has shape (*k, n)
    Returns: Array with shape (*k, n, n) and the elements of m on the diagonals.
    """
```

NeilGirdhar:
Fixes #136

### DIFF
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -12,7 +12,7 @@ from typing import cast
 from ._at import at
 from ._utils import _compat, _helpers
 from ._utils._compat import array_namespace, is_jax_array
-from ._utils._helpers import asarrays
+from ._utils._helpers import asarrays, ndindex
 from ._utils._typing import Array
 
 __all__ = [
@@ -236,7 +236,7 @@ def create_diagonal(
     Parameters
     ----------
     x : array
-        A 1-D array.
+        An array having shape ``(*batch_dims, k)``.
     offset : int, optional
         Offset from the leading diagonal (default is ``0``).
         Use positive ints for diagonals above the leading diagonal,
@@ -247,7 +247,8 @@ def create_diagonal(
     Returns
     -------
     array
-        A 2-D array with `x` on the diagonal (offset by `offset`).
+        An array having shape ``(*batch_dims, k+abs(offset), k+abs(offset))`` with `x`
+        on the diagonal (offset by `offset`).
 
     Examples
     --------
@@ -270,18 +271,21 @@ def create_diagonal(
     if xp is None:
         xp = array_namespace(x)
 
-    if x.ndim != 1:
-        err_msg = "`x` must be 1-dimensional."
+    if x.ndim == 0:
+        err_msg = "`x` must be at least 1-dimensional."
         raise ValueError(err_msg)
-    n = x.shape[0] + abs(offset)
-    diag = xp.zeros(n**2, dtype=x.dtype, device=_compat.device(x))
+    batch_dims = x.shape[:-1]
+    n = x.shape[-1] + abs(offset)
+    diag = xp.zeros((*batch_dims, n**2), dtype=x.dtype, device=_compat.device(x))
 
-    start = offset if offset >= 0 else abs(offset) * n
-    stop = min(n * (n - offset), diag.shape[0])
-    step = n + 1
-    diag = at(diag)[start:stop:step].set(x)
-
-    return xp.reshape(diag, (n, n))
+    target_slice = slice(
+        offset if offset >= 0 else abs(offset) * n,
+        min(n * (n - offset), diag.shape[-1]),
+        n + 1,
+    )
+    for index in ndindex(*batch_dims):
+        diag = at(diag)[(*index, target_slice)].set(x[(*index, slice(None))])
+    return xp.reshape(diag, (*batch_dims, n, n))
 
 
 def expand_dims(
--- a/src/array_api_extra/_lib/_utils/_helpers.py
+++ b/src/array_api_extra/_lib/_utils/_helpers.py
@@ -3,6 +3,7 @@
 # https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
 from __future__ import annotations
 
+from collections.abc import Generator
 from types import ModuleType
 from typing import cast
 
@@ -175,3 +176,26 @@ def asarrays(
         xa, xb = xp.asarray(a), xp.asarray(b)
 
     return (xb, xa) if swap else (xa, xb)
+
+
+def ndindex(*x: int) -> Generator[tuple[int, ...]]:
+    """
+    Generate all N-dimensional indices for a given array shape.
+
+    Given the shape of an array, an ndindex instance iterates over the N-dimensional
+    index of the array. At each iteration a tuple of indices is returned, the last
+    dimension is iterated over first.
+
+    This has an identical API to numpy.ndindex.
+
+    Parameters
+    ----------
+    *x : int
+        The shape of the array.
+    """
+    if not x:
+        yield ()
+        return
+    for i in ndindex(*x[:-1]):
+        for j in range(x[-1]):
+            yield *i, j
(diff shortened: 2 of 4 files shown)


NeilGirdhar:
I'm not entirely sure what I should do about the numpy dependency.  Do I re-implement `ndindex` somewhere?

Also, I'm getting a strange error when trying to run the tests about a bad device.

lucascolley:
yes, we will want to remove the NumPy dependency. I don't know how easy it is to make `ndindex` agnostic.

> Also, I'm getting a strange error when trying to run the tests about a bad device.

`np.ndindex` will be producing arrays to the default NumPy device, so you need to be careful to coerce to the correct device when that is used (things are probably going wrong in the `at` call). But this problem should be subsumed in removing the NumPy dependency.

NeilGirdhar:
Okay, I implemented `ndindex`, but I guess it will be slower than Numpy's version since mine is in Python.

Anyway, I'm still getting some inscrutable errors, if you have any ideas?

NeilGirdhar:
Ready for your review :smile:

lucascolley:
@mdhaber has `ndindex` come up in any of your work on vectorising things?

NeilGirdhar:
What do you mean?  I just need it for this `create_diagonal`.

lucascolley:
I'm just wondering whether this is an instance of a problem that has appeared elsewhere

NeilGirdhar:
Whenever I need `ndindex`, I use the one from Numpy.  I've never been reluctant to import Numpy, so I would not use a version from xpx.

lucascolley:
Sure, I was just asking Matt if anything like this had come up in the context of array-agnostic code yet

NeilGirdhar:
Oh sorry!   I didn't see the tag.  I haven't been able to sleep tonight and I'm just tired.

lucascolley:
no worries!

mdhaber:
Yes, I used it in the decorator that adds batch support in `linalg`. For the decorator used in stats, I considered using `ndindex` before going with `apply_along_axis`, but if I were to rewrite it I'd just use `ndindex`. Both of these are just for NumPy right now, though.

I've thought about having a heuristic in `marray` that decides whether to loop over slices and compress (remove masked elements from) each slice before performing the operation. This can be faster than the usual thing (say, `data[mask] = 0` in `sum`) when there are fewer slices and they are very big. But for that, I assumed I would reshape to have just one batch dimension at the front, loop over that single axis, and then reshape the result. But by "reshape" I really mean `moveaxis` + `reshape`, and `ndindex` would save the trouble of `reshape`.

lucascolley:
Makes sense, thanks! In this case, does the approach of making `ndindex` agnostic seem good to you?

NeilGirdhar:
@lucascolley Would you like me to do anything else with this PR?

lucascolley:
I just need to find time to review this. @j-bowhay would you like to take a look?

NeilGirdhar:
Sounds good.  I added a lot more tests.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -172,7 +172,7 @@ def create_diagonal(
     Parameters
     ----------
     x : array
-        A 1-D array.
+        An array having shape (*broadcast_dims, k).
```
j-bowhay, line 175: Nit, but at least in SciPy we'd call this `batch_dims`. I'll let @lucascolley decide what language he'd prefer

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -175,3 +176,26 @@ def asarrays(
         xa, xb = xp.asarray(a), xp.asarray(b)
 
     return (xb, xa) if swap else (xa, xb)
+
+
+def ndindex(*x: int) -> Generator[tuple[int, ...]]:
```
j-bowhay, line 181: It would be good to add a test to compare against numpy

tests/test_funcs.py
```diff
@@ -193,9 +194,18 @@ def test_0d(self, xp: ModuleType):
...
+    @pytest.mark.parametrize("shape", [(0,), (10,), (0, 1), (1, 0), (0, 0), (4, 2, 1),
+                                       (1, 1, 7), (0, 0, 1), (3, 2, 4, 5)])
+    def test_nd(self, xp: ModuleType, shape: tuple[int, ...]):
+        rng = np.random.default_rng(2347823)
+        b = xp.asarray(rng.integers((1 << 64) - 1, size=shape, dtype=np.uint64),
+                       dtype=xp.uint64)
+        c = create_diagonal(b)
+        zero = xp.zeros((), dtype=xp.uint64)
+        assert c.shape == (*b.shape, b.shape[-1])
+        for i in ndindex(*c.shape):
```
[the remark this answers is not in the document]
NeilGirdhar: If we test ndindex, would that be good enough?  I'm worried that the above approach is hard to understand?

src/array_api_extra/_lib/_funcs.py
NeilGirdhar: Sure, let me know

src/array_api_extra/_lib/_utils/_helpers.py
NeilGirdhar: Done.

tests/test_funcs.py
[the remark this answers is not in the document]
j-bowhay: Here's an example of pattern
https://github.com/scipy/scipy/blob/f7da7b74bcf297aa812e14f284b49e6be21e3490/scipy/linalg/tests/test_special_matrices.py#L621
Let me know what you think if you still think it would be hard to understand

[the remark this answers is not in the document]
NeilGirdhar: For me, that's a lot harder to understand.  But maybe you're more familiar with this pattern.

Consider what we have:
```python
        for i in ndindex(*c.shape):   # Iterate over everything
            xp_assert_equal(c[i], b[i[:-1]] if i[-2] == i[-1] else zero)  # Check that diagonal elements match.
```
versus reshaping, calling create_diagonal on one-dimensional slices and then checking that these match.  This trades dependency on ndarray working correctly (easily tested) for dependency on one-dimensional create-diagonal, which is what we're trying to test.

I think this might have made sense in scipy where they already had non-batching code working and then added batching code later.  The `test_nd` is designed, it even supplants `test_1d`.

NeilGirdhar:
(I renamed the test functions to make it a bit clearer why we have each test.)

src/array_api_extra/_lib/_funcs.py
```diff
@@ -183,7 +183,8 @@ def create_diagonal(
     Returns
     -------
     array
-        A 2-D array with `x` on the diagonal (offset by `offset`).
+        An array having shape (*broadcast_dims, k+abs(offset), k+abs(offset)) with `x`
```
j-bowhay, line 186:
```suggestion
        An array having shape ``(*broadcast_dims, k+abs(offset), k+abs(offset))`` with `x`
```

src/array_api_extra/_lib/_funcs.py
j-bowhay: Either way possibly should be monospace ?
```suggestion
        An array having shape ``(*broadcast_dims, k)``.
```

j-bowhay (approved):
LGTM modulo the one query for @lucascolley

NeilGirdhar:
@j-bowhay Thank you for much for you thorough and speedy review.  We'll wait on Lucas's decision about the term to use.

src/array_api_extra/_lib/_funcs.py
lucascolley: No strong opinion either way, so let's go for `batch_dims` to match SciPy

NeilGirdhar: Already done!  :smile:

lucascolley:
looks like there is a merge conflict

NeilGirdhar:
Should be fixed now

NeilGirdhar:
> I will merge this after putting out a micro release.

Awesome thanks!  Just curious, why not merge then release?

lucascolley:
thanks again @NeilGirdhar !

PR #138, opened 2025-02-08
DEV: Add editorconfig

NeilGirdhar:


lucascolley (commented):
I think some of these rules are different to what is enforced by the linter, e.g. the max line length for py files is 88.

What is the reason for wanting this in the repo? Can we not just `.gitignore` it instead?

NeilGirdhar:
It makes the editor use your settings for things like line wrap.

lucascolley (approved):
alright, LGTM when CI is happy!

lucascolley:
thx!

PR #141, opened 2025-02-10
ENH: `apply_where` (migrate `lazywhere` from scipy)

### LINKED ISSUE #14: ENH: add `lazywhere`
@mdhaber do you think `lazywhere` would be in scope for this library? It seems to me like it would be useful beyond SciPy?

crusaderky:
Closes #14.
`scipy._lib._utils.lazywhere`, with revised UI and added support for jax.jit and Dask.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ numpydoc = ">=1.8.0,<2"
 array-api-strict = "*"
 numpy = "*"
 pytest = "*"
+hypothesis = "*"
 dask-core = "*"  # No distributed, tornado, etc.
 # NOTE: don't add jax, pytorch, sparse, cupy here
 # as they slow down mypy and are not portable across target OSs
@@ -79,6 +80,7 @@ lint = { depends-on = ["pre-commit", "pylint", "mypy", "pyright"] }
 [tool.pixi.feature.tests.dependencies]
 pytest = ">=6"
 pytest-cov = ">=3"
+hypothesis = "*"
 array-api-strict = "*"
 numpy = "*"
 
@@ -232,6 +234,10 @@ reportMissingTypeStubs = false
 reportUnreachable = false
 # ruff handles this
 reportUnusedParameter = false
+# cyclic imports inside function bodies
+reportImportCycles = false
+# PyRight can't trace types in lambdas
+reportUnknownLambdaType = false
 
 executionEnvironments = [
     { root = "tests", reportPrivateUsage = false },
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import math
 import warnings
 from collections.abc import Callable, Sequence
-from types import ModuleType
+from types import ModuleType, NoneType
 from typing import cast, overload
 
 from ._at import at
@@ -54,7 +54,7 @@ def apply_where(  # type: ignore[no-any-explicit,no-any-decorated] # numpydoc ig
     f1: Callable[..., Array],
     /,
     *,
-    fill_value: Array | int | float | complex | bool,
+    fill_value: Array | complex,
     xp: ModuleType | None = None,
 ) -> Array: ...
 
@@ -66,7 +66,7 @@ def apply_where(  # type: ignore[no-any-explicit] # numpydoc ignore=PR01,PR02
     f2: Callable[..., Array] | None = None,
     /,
     *,
-    fill_value: Array | int | float | complex | bool | None = None,
+    fill_value: Array | complex | None = None,
     xp: ModuleType | None = None,
 ) -> Array:
     """
@@ -115,11 +115,13 @@ def apply_where(  # type: ignore[no-any-explicit] # numpydoc ignore=PR01,PR02
 
     Examples
     --------
+    >>> import array_api_strict as xp
+    >>> import array_api_extra as xpx
     >>> a = xp.asarray([5, 4, 3])
     >>> b = xp.asarray([0, 2, 2])
     >>> def f(a, b):
     ...     return a // b
-    >>> apply_where(b != 0, (a, b), f, fill_value=xp.nan)
+    >>> xpx.apply_where(b != 0, (a, b), f, fill_value=xp.nan)
     array([ nan,  2., 1.])
     """
     # Parse and normalize arguments
@@ -129,12 +131,12 @@ def apply_where(  # type: ignore[no-any-explicit] # numpydoc ignore=PR01,PR02
     args_ = list(args) if isinstance(args, tuple) else [args]
     del args
 
-    xp = array_namespace(cond, *args_) if xp is None else xp
+    xp = array_namespace(cond, fill_value, *args_) if xp is None else xp
 
-    if getattr(fill_value, "ndim", 0):
-        cond, fill_value, *args_ = xp.broadcast_arrays(cond, fill_value, *args_)
-    else:
+    if isinstance(fill_value, int | float | complex | NoneType):
         cond, *args_ = xp.broadcast_arrays(cond, *args_)
+    else:
+        cond, fill_value, *args_ = xp.broadcast_arrays(cond, fill_value, *args_)
 
     if is_dask_namespace(xp):
         meta_xp = meta_namespace(cond, fill_value, *args_, xp=xp)
@@ -161,10 +163,10 @@ def _apply_where(  # type: ignore[no-any-explicit]  # numpydoc ignore=PR01,RT01
 
     if f2 is None:
         dtype = xp.result_type(temp1, fill_value)
-        if getattr(fill_value, "ndim", 0):
-            out = xp.astype(fill_value, dtype, copy=True)
-        else:
+        if isinstance(fill_value, int | float | complex):
             out = xp.full_like(cond, dtype=dtype, fill_value=fill_value)
+        else:
+            out = xp.astype(fill_value, dtype, copy=True)
     else:
         ncond = ~cond
         temp2 = f2(*(arr[ncond] for arr in args))
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -5,17 +5,23 @@ from __future__ import annotations
 
 import math
 import warnings
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from types import ModuleType
-from typing import cast
+from typing import cast, overload
 
 from ._at import at
 from ._utils import _compat, _helpers
-from ._utils._compat import array_namespace, is_jax_array
-from ._utils._helpers import asarrays, eager_shape, ndindex
+from ._utils._compat import (
+    array_namespace,
+    is_dask_namespace,
+    is_jax_array,
+    is_jax_namespace,
+)
+from ._utils._helpers import asarrays, eager_shape, meta_namespace, ndindex
 from ._utils._typing import Array
 
 __all__ = [
+    "apply_where",
     "atleast_nd",
     "broadcast_shapes",
     "cov",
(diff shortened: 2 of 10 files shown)


src/array_api_extra/_lib/_funcs.py
```diff
@@ -28,6 +36,168 @@
...
+    *args: Array,
+    fill_value: Array | int | float | complex | bool | None,
+    dtype: DType,
+    xp: ModuleType,
+) -> Array:
+    """Helper of `apply_where`. On Dask, this runs on a single chunk."""
+
+    if is_jax_namespace(xp):
+        # jax.jit does not support assignment by boolean mask
+        return xp.where(cond, f1(*args), f2(*args) if f2 is not None else fill_value)
```
crusaderky, line 184: JAX-on-dask is currently unsupported by Dask. This is here and not much higher above only for this reason.
We'll need to add explicit unit tests when it lands in the future.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -28,6 +36,168 @@
...
+    cond, *args = xp.broadcast_arrays(cond, *args)  # pyright: ignore[reportAssignmentType]
+    temp1 = f1(*(arr[cond] for arr in args))
+
+    if f2 is None:
+        out = xp.full(cond.shape, fill_value=fill_value, dtype=dtype, device=device)
+    else:
+        ncond = ~cond
+        temp2 = f2(*(arr[ncond] for arr in args))
+        out = xp.empty(cond.shape, dtype=dtype, device=device)
+        out = at(out, ncond).set(temp2)
```
crusaderky, line 196: JAX doesn't benefit from this, but Sparse will (eventually)

src/array_api_extra/_lib/_funcs.py
```diff
@@ -28,6 +36,168 @@
...
+    xp: ModuleType | None = None,
+) -> Array: ...
+
+
+def apply_where(  # type: ignore[no-any-explicit,misc] # numpydoc ignore=PR01,PR02
+    cond: Array,
+    f1: Callable[..., Array],
+    f2: Callable[..., Array] | Array,
+    /,
+    *args: Array,
```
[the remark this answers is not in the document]
crusaderky: > There should not be any positional arguments after `*args`

There aren't. All arguments after a variadic argument are always keyword-only.

> Also, if f2 is given positionally here, fill_value should not even be present, just like in the first overload f2 isn't present?

Correct, the two valid signatures are the ones in the typing annotations. This is enforced at runtime.
`cond`, `f1`, `f2`, and `*args` are position-only; `fill_value` is keyword-only.
There are exclusively two possible ways to call this function:
```python
apply_where(cond, f1, f2, x1, x2, x3, ...)
apply_where(cond, f1, x1, x2, x3, ..., fill_value=123)
```
I felt that putting `f1, f2` next to each other was more intuitive coming from the syntax of `xp.where`. I can change it if you feel strongly about it though.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -28,6 +36,168 @@
...
+            raise ValueError(msg)
+    del f2
+    if not args:
+        msg = "Must give at least one input array."
+        raise TypeError(msg)
+
+    xp = array_namespace(cond, *args) if xp is None else xp
+
+    # Determine output dtype
+    metas = [get_meta(arg, xp=xp) for arg in args]
```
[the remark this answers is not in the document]
crusaderky: Yes, Dask is the only one that needs to know the output dtype ahead of time. I will refactor it.

src/array_api_extra/_lib/_funcs.py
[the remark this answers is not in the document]
rgommers: > `fill_value` is keyword-only.

So adding `*, `to communicate that would be useful, right?

> I felt that putting `f1, f2` next to each other was more intuitive coming from the syntax of `xp.where`.

I think it's fine to leave it, I agree f1/f2 together makes sense. The overloads plus `*args` make the signature look not so great though, compared to the original. Not sure if there's a better option - maybe `arrays, *, fill_value` is more natural.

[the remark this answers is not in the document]
crusaderky: > > `fill_value` is keyword-only.
> 
> So adding `*, `to communicate that would be useful, right?

You can't; the `*` is already in `*args`.
```python
In [1]: def f(*args, *, kwonly): pass
  Cell In[17], line 1
    def f(*args, *, kwonly): pass
                 ^
SyntaxError: * argument may appear only once
```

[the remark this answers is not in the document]
crusaderky: Changed args to an explicit tuple. I found myself writing `*(a, b)` every time anyway.

pyproject.toml
```diff
@@ -191,10 +191,6 @@ markers = [
 
 [tool.coverage]
 run.source = ["array_api_extra"]
-report.exclude_also = [
-    '\.\.\.',
-    'if TYPE_CHECKING:',
-]
```
crusaderky, line 197: This was causing coverage to incorrectly skip vast sections of code

src/array_api_extra/_lib/_funcs.py
```diff
@@ -385,12 +542,15 @@ def isclose(
...
+        mxp = meta_namespace(a, b, xp=xp)
+        out = apply_where(
             xp.isinf(a) | xp.isinf(b),
-            xp.isinf(a) & xp.isinf(b) & (xp.sign(a) == xp.sign(b)),
+            lambda a, b: mxp.isinf(a) & mxp.isinf(b) & (mxp.sign(a) == mxp.sign(b)),  # pyright: ignore[reportUnknownArgumentType]
             # Note: inf <= inf is True!
-            xp.abs(a - b) <= (atol + rtol * xp.abs(b)),
+            lambda a, b: mxp.abs(a - b) <= (atol + rtol * mxp.abs(b)),  # pyright: ignore[reportUnknownArgumentType]
+            (a, b),
+            xp=xp,
```
crusaderky, line 553: If you forget about the meta-namespace at the moment it is impossible to trigger a failure in unit tests.
This will become a problem when Dask will eventually allow wrapping around arbitrary Array API compatible namespaces. At the moment, these dask ufuncs are just thin wrappers around the corresponding numpy ufuncs, and all the namespaces that dask _does_ wrap around (numpy, cupy, and sparse) all support `__array_ufunc__`.
In other words, `da.isinf(cp.asarray(0.))` works thanks to `__array_ufunc__`, but is illegal in the Array API standard.

This pattern repeats itself many, many times in scipy. At the moment the handful of cases that are array API-aware were just using `xp.divide` and I worked around the problem by replacing it with `operator.truediv`. But if you look at scipy.stats in https://github.com/scipy/scipy/pull/22557 you'll find a ton of calls to `np.` functions inside the lambdas.

I see two possible ways forward:
1. make `meta_namespace` public API.
_like:_ explicit is better than implicit
_dislike:_ very verbose

2. add signature magic to `apply_where`; if f1 or f2 accept a keyword argument called "xp", pass to it the meta-namespace:
```python
        out = apply_where(
            ...,  # cond
            lambda a, b, xp: xp.isinf(a) & xp.isinf(b) & (xp.sign(a) == xp.sign(b)),
            lambda a, b, xp: xp.abs(a - b) <= (atol + rtol * xp.abs(b)),
            (a, b)
        )
```
_like:_ synthetic; no need for helper functions
_dislike:_ obscure functionality which needs to be commented every time; otherwise unwary maintainers will break it by trying to simplify it. (this negates its compactness benefit)

tests/conftest.py
crusaderky, line 1: All changes to this module are identical to https://github.com/data-apis/array-api-extra/pull/86

src/array_api_extra/_lib/_funcs.py
```diff
@@ -29,6 +35,157 @@
...
+
+
+@overload
+def apply_where(  # type: ignore[no-any-explicit,no-any-decorated] # numpydoc ignore=GL08
+    cond: Array,
+    f1: Callable[..., Array],
+    args: Array | tuple[Array, ...],
+    /,
+    *,
+    fill_value: Array | int | float | complex | bool,
```
mdhaber, line 57: I think I'd support the overloading the third positional argument could be used for either `fill_value` or `f2` because they are conceptually similar. But I am getting confused in gh-22557 by the third positional argument being `args` sometimes.

crusaderky: I think that would be problematic. `apply_where(cond, func, x, y)` looks extremely obscure to me. It's calling neither `f(x, y)` nor `f(x) if cond else y` as one would intuitively read it; it's calling `f(y) if cond else x`.
By contrast, `apply_where(cond, func, y, fill_value=x)` looks unambiguous, also to a distracted reader.

tests/test_helpers.py
```diff
@@ -155,3 +158,21 @@ def test_numpy_generics(self, dtype: type):
...
+
+    def test_dask_metas(self, da: ModuleType):
+        cp = pytest.importorskip("cupy")
+        cp_compat = array_namespace(cp.empty(0))
+        args = None, da.from_array(cp.asarray(0)), 1
+        assert meta_namespace(*args) is cp_compat
+
+    def test_xp(self, xp: ModuleType):
+        args = None, xp.asarray(0), 1
+        assert meta_namespace(*args, xp=xp) in (xp, np_compat)
```
lucascolley, line 178: Is this just the same as above where we have `np_compat if library is Backend.DASK else xp` ? If so, no need to change it, just wondering if I'm missing something.

lucascolley (commented):
partial review (all apart from `apply_where` itself and its tests)

src/array_api_extra/_lib/_funcs.py
```diff
@@ -29,6 +35,148 @@
...
+    """Helper of `apply_where`. On Dask, this runs on a single chunk."""
+
+    if is_jax_namespace(xp):
+        # jax.jit does not support assignment by boolean mask
+        return xp.where(cond, f1(*args), f2(*args) if f2 is not None else fill_value)
+
+    temp1 = f1(*(arr[cond] for arr in args))
+
+    if f2 is None:
+        # TODO remove asarrays once all backends support Array API 2024.12
```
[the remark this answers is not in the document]
crusaderky: > Can `asarrays` be removed now that `array-api-compat` has 2024.12 support? 

Yep. Good catch.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -29,6 +35,146 @@
...
+    On Dask, `f1` and `f2` are applied to the individual chunks and should use functions
+    from the namespace of the chunks.
+
+    Examples
+    --------
+    >>> a = xp.asarray([5, 4, 3])
+    >>> b = xp.asarray([0, 2, 2])
+    >>> def f(a, b):
+    ...     return a // b
+    >>> apply_where(b != 0, (a, b), f, fill_value=xp.nan)
```
lucascolley, line 122: Perhaps?
```
    >>> import array_api_strict as xp
    >>> import array_api_extra as xpx
    ....
    xpx.apply_where
```

src/array_api_extra/_lib/_funcs.py
```diff
@@ -29,6 +35,146 @@
...
+
+    if is_jax_namespace(xp):
+        # jax.jit does not support assignment by boolean mask
+        return xp.where(cond, f1(*args), f2(*args) if f2 is not None else fill_value)
+
+    temp1 = f1(*(arr[cond] for arr in args))
+
+    if f2 is None:
+        dtype = xp.result_type(temp1, fill_value)
+        if getattr(fill_value, "ndim", 0):
```
lucascolley, line 164: nit
```suggestion
        if getattr(fill_value, "ndim", False):
```

lucascolley (commented):
thanks @crusaderky, LGTM once open comments are resolved!

src/array_api_extra/_lib/_funcs.py
```diff
@@ -29,6 +35,146 @@
...
+    # Parse and normalize arguments
+    if (f2 is None) == (fill_value is None):
+        msg = "Exactly one of `fill_value` or `f2` must be given."
+        raise TypeError(msg)
+    args_ = list(args) if isinstance(args, tuple) else [args]
+    del args
+
+    xp = array_namespace(cond, *args_) if xp is None else xp
+
+    if getattr(fill_value, "ndim", 0):
```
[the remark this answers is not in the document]
crusaderky: I like it more as it is. It emphasizes that an array with ndim=0 should behave the same as a Python scalar.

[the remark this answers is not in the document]
lucascolley: okay, sure

lucascolley (approved):
would you like to move the discussion at https://github.com/data-apis/array-api-extra/pull/141#discussion_r1961964801 to an issue somewhere?

src/array_api_extra/_lib/_funcs.py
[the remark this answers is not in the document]
crusaderky: Just realized that full_like doesn't accept a scalar array fill_value. Fixing it.

crusaderky:
@lucascolley all comments have been addressed

PR #145, opened 2025-02-20
TST: refactor test_utils

crusaderky:
- rename test_utils to test_helpers, in line with changes with the tested module
- refactor tests for asarrays to use a namespace


lucascolley (approved):
thanks!

PR #156, opened 2025-03-14
MAINT: Array API 2024.12 typing nits

crusaderky:


crusaderky:
@lucascolley ready for review and merge

src/array_api_extra/_lib/_funcs.py
```diff
@@ -385,6 +385,9 @@ def isclose(
 ) -> Array:  # numpydoc ignore=PR01,RT01
     """See docstring in array_api_extra._delegation."""
     a, b = asarrays(a, b, xp=xp)
+    if TYPE_CHECKING:  # Hack around pyright bug  # pragma: no cover
```
lucascolley, line 388: What is the pyright bug?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,8 +375,8 @@ def expand_dims(
 
 
 def isclose(
-    a: Array,
-    b: Array,
+    a: Array | complex,
+    b: Array | complex,
```
lucascolley, line 379: Do we want to update the docstrings to clarify that scalars are accepted?

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -96,16 +101,16 @@ def mean(
     return xp.mean(x, axis=axis, keepdims=keepdims)
 
 
-def is_python_scalar(x: object) -> bool:  # numpydoc ignore=PR01,RT01
+def is_python_scalar(x: object) -> TypeIs[complex]:  # numpydoc ignore=PR01,RT01
     """Return True if `x` is a Python scalar, False otherwise."""
     # isinstance(x, float) returns True for np.float64
     # isinstance(x, complex) returns True for np.complex128
-    return isinstance(x, int | float | complex | bool) and not is_numpy_array(x)
+    return isinstance(x, int | float | complex) and not is_numpy_array(x)
```
lucascolley, line 108: `bool` check was redundant?

src/array_api_extra/_lib/_funcs.py
crusaderky: given
```python
def f(x: Array | complex) -> Array: ...

x: Array | complex
x = f(x)
```
when you reach the end of the snippet, mypy correctly understands that x can only be Array, whereas pyright still believes it could be complex.

src/array_api_extra/_lib/_utils/_helpers.py
crusaderky: it's a subclass of int

src/array_api_extra/_lib/_funcs.py
crusaderky: I've now reviewed all

src/array_api_extra/_lib/_funcs.py
lucascolley: ah, so is it about reusing variable names? Is there an upstream issue we can link to?

src/array_api_extra/_lib/_utils/_helpers.py
lucascolley, line 108:
```suggestion
    # bool is a subclass of int
    return isinstance(x, int | float | complex) and not is_numpy_array(x)
```

src/array_api_extra/_lib/_funcs.py
lucascolley: given the upstream answer, what would you like to do here? Use different variable names? Keep the current workarounds but remove the `FIXME`?

crusaderky: I've replaced `Array` with a fully fledged Protocol. As it would vastly enlarge the scope of this PR, I instead opened a follow-up that reverts these three hacks: https://github.com/data-apis/array-api-extra/pull/159

lucascolley (approved):
thanks @crusaderky !

PR #158, opened 2025-03-14
ENH: `lazy_xp_function` namespaces support

crusaderky:
Working in scipy I realized that forcing the developer who writes the tests to `import <names> from <module>` can be extremely cumbersome - see for example `scipy.stats.tests.test_stats`.

This PR allows simply calling 
```python

import <module>
lazy_xp_modules = [<module>]

def test_<funcname>(xp):
    <module>.<funcname>(...)
```

### DIFF
--- a/src/array_api_extra/testing.py
+++ b/src/array_api_extra/testing.py
@@ -56,9 +56,8 @@ def lazy_xp_function(  # type: ignore[no-any-explicit]
     """
     Tag a function to be tested on lazy backends.
 
-    Tag a function, which must be imported in the test module globals, so that when any
-    tests defined in the same module are executed with ``xp=jax.numpy`` the function is
-    replaced with a jitted version of itself, and when it is executed with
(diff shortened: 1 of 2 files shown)


crusaderky:
@lucascolley ready for review and merge

src/array_api_extra/testing.py
```diff
@@ -121,19 +120,59 @@ def test_myfunc(xp):
 
     Notes
     -----
-    A test function can circumvent this monkey-patching system by calling `func` as an
-    attribute of the original module. You need to sanitize your code to make sure this
-    does not happen.
+    In order for this tag to be effective, the test function must be imported into the
+    test module globals without namespace; alternatively its namespace must be declared
```
lucascolley, line 124: Perhaps?
```suggestion
    test module globals without its namespace; alternatively its namespace must be declared
```

lucascolley (approved):
thanks @crusaderky, LGTM!!

PR #169, opened 2025-03-19
TST: NumPy 1.x support

crusaderky:


src/array_api_extra/_lib/_funcs.py
```diff
@@ -565,12 +565,11 @@ def isclose(
...
         return xp.abs(a - b) <= atol
 
-    try:
-        nrtol = xp.asarray(int(1.0 / rtol), dtype=b.dtype)
-    except OverflowError:
-        # rtol * max_int(dtype) < 1, so it's inconsequential
+    # Don't rely on OverflowError, as it is not guaranteed by the Array API.
+    nrtol = int(1.0 / rtol)
+    if nrtol > xp.iinfo(b.dtype).max:
+        # rtol * max_int < 1, so it's inconsequential
```
crusaderky, line 571: Impacts array-api-strict wrapping around numpy 1.26

lucascolley:
> Out of scope: we need to start a conversation about how array-api-compat, array-api-extra, and scipy are going to support obsolete versions of backends other than NumPy.

Yeah, I think we are still at a stage where we can require that only the latest versions work. But we'll get to this point eventually.

lucascolley:
I linked the issue suggesting that we test with oldest SPEC 0 NumPy. Pros/cons of that vs. the approach of `<2` here? Probably almost the same in practice.

crusaderky:
> I linked the issue suggesting that we test with oldest SPEC 0 NumPy. Pros/cons of that vs. the approach of `<2` here? Probably almost the same in practice.

That makes sense. Now testing vs. 1.25.0.

lucascolley (approved):
thanks @crusaderky !

lucascolley:
we should probably change the `numpy1` env to test NumPy 1.22, as that is now the oldest supported by scikit-learn. Hopefully SPEC 0 will evolve to eventually converge with scikit-learn somehow, but there is no need to be more aggressive in the meantime.

PR #170, opened 2025-03-20
MAINT: drop Python 3.9 informal support

### LINKED ISSUE #167: MAINT/DEV: drop Python 3.9 shims
@crusaderky would you prefer to write or review a PR for this?

crusaderky:
Closes #167

### DIFF
--- a/src/array_api_extra/_lib/_at.py
+++ b/src/array_api_extra/_lib/_at.py
@@ -1,13 +1,12 @@
 """Update operations for read-only arrays."""
 
-# https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
 from __future__ import annotations
 
 import operator
 from collections.abc import Callable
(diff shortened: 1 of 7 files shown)


src/array_api_extra/_lib/_at.py
```diff
@@ -1,13 +1,12 @@
 """Update operations for read-only arrays."""
 
-# https://github.com/scikit-learn/scikit-learn/pull/27910#issuecomment-2568023972
 from __future__ import annotations
```
lucascolley, line 3: Can we not drop this?

crusaderky: No, because delayed annotations won't be the default until a release TBD: https://docs.python.org/3/library/__future__.html#module-contents

As a result any module that uses `if TYPE_CHECKING:` or recursive annotations will need it indefinitely.

PR #177, opened 2025-03-25
BUG: Fix isclose multidevice for equal_nan=True

lithomas1:
``xp.asarray(True)`` was missing a device argument, which causes array-api-strict with a device argument to fail in where.

Since as of array api 2024.12 we can pass in scalars directly to where, I've replaced asarray(True) with just True.


tests/conftest.py
```diff
@@ -131,7 +131,11 @@ def xp(
         # suppress unused-ignore to run mypy in -e lint as well as -e dev
         jax.config.update("jax_enable_x64", True)  # type: ignore[no-untyped-call,unused-ignore]
 
-    return xp
+    device = None
+    if library == Backend.ARRAY_API_STRICT_DEVICE1:
+        import array_api_strict
+        device = array_api_strict.Device("device1")
+    return xp, device
```
lithomas1, line 138: I was trying to add multi-device testing here similar to the way scikit-learn does it.
https://github.com/scikit-learn/scikit-learn/pull/30090/files

I don't know if this is the right approach, though.
I didn't fix all the tests - I wanted to get some feedback before making a very invasive change like this (I'd need to migrate all the tests).

lithomas1:
@lucascolley 
Do you mind taking a quick look at this PR?

lucascolley (commented):
we already have 

https://github.com/data-apis/array-api-extra/blob/acbdf72849accc99c790adfd66e87f50fe311096/tests/conftest.py#L148-L161

it looks like we just forgot to add a `test_device` for `isclose`, see e.g. https://github.com/data-apis/array-api-extra/blob/acbdf72849accc99c790adfd66e87f50fe311096/tests/test_funcs.py#L435-L437

lithomas1:
Wow, totally missed that!

Thanks, that's a lot cleaner.

lucascolley (approved):
thanks Thomas!

lithomas1:
Sorry, sent in a fix!

lucascolley:
Unfortunately the `xp_assert` infra is not ready for array-api-strict arrays which can't be converted to NumPy arrays.

I'd be happy for you to modify the test to be just a device check with `get_device`. Alternatively, you could look into making the `xp_assert` infra work, but I'm not sure what the best way to go about doing that is.

lithomas1:
> I'd be happy for you to modify the test to be just a device check with `get_device`. Alternatively, you could look into making the `xp_assert` infra work, but I'm not sure what the best way to go about doing that is.

I added a conversion to the CPU device in array-api-strict which is what scikit-learn does as well I think.
Let me know if that approach works for you.

lithomas1:
Thanks for the review and for fixing typing!

PR #179, opened 2025-03-25
ENH/TST: tougher restrictions on array_api_strict

crusaderky:
- Run all tests with array-api-strict with all optional capabilities disabled
- Add `nunique` support for unknown backends without `unique_counts` - for example, `marray` wrapping around JAX.
- Add `apply_where` support for unknown backends without boolean index assignment - again for example `marray` wrapping around JAX (but see https://github.com/jax-ml/jax/issues/27418) 
- Add `apply_where` support for sparse arrays

# Upstream issues
- https://github.com/jax-ml/jax/issues/27418
- https://github.com/data-apis/array-api-compat/pull/290
- https://github.com/mdhaber/marray/issues/99

### DIFF
--- a/src/array_api_extra/_lib/_utils/_helpers.py
+++ b/src/array_api_extra/_lib/_utils/_helpers.py
@@ -12,7 +12,9 @@ from ._compat import (
     array_namespace,
     is_array_api_obj,
     is_dask_namespace,
+    is_jax_namespace,
     is_numpy_array,
+    is_pydata_sparse_namespace,
 )
 from ._typing import Array
 
@@ -23,6 +25,7 @@ if TYPE_CHECKING:  # pragma: no cover
 
 __all__ = [
     "asarrays",
+    "capabilities",
     "eager_shape",
     "in1d",
     "is_python_scalar",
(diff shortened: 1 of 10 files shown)


lucascolley (commented):
high-level question before reviewing this: what is the motivation for keeping both `ARRAY_API_STRICT` and `ARRAY_API_STRICTEST`, rather than just only using the strictest version?

crusaderky:
> high-level question before reviewing this: what is the motivation for keeping both `ARRAY_API_STRICT` and `ARRAY_API_STRICTEST`, rather than just only using the strictest version?

ARRAY_API_STRICT follows different, more performant code paths. You can see it both in `nunique` and in `apply_where`.

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -270,3 +273,33 @@ def meta_namespace(
     # Quietly skip scalars and None's
     metas = [cast(Array | None, getattr(a, "_meta", None)) for a in arrays]
     return array_namespace(*metas)
+
+
+def capabilities(xp: ModuleType) -> dict[str, int]:
```
lucascolley, line 278: can we add a TODO to upstream this to array-api-compat?

tests/test_funcs.py
```diff
@@ -819,8 +829,29 @@ def test_empty(self, xp: ModuleType):
...
+        xp_assert_equal(nunique(a), xp.asarray(1))
+
+    def test_all_equal(self, xp: ModuleType):
+        a = xp.asarray([123, 123, 123])
+        xp_assert_equal(nunique(a), xp.asarray(1))
+
+    @pytest.mark.xfail_xp_backend(Backend.DASK, reason="No equal_nan kwarg in unique")
+    @pytest.mark.xfail_xp_backend(
+        Backend.SPARSE, reason="Non-compliant equal_nan=True behaviour"
+    )
```
lucascolley, line 843: upstream issue?

lucascolley (approved):
thanks Guido, looks pretty much there!

src/array_api_extra/_lib/_utils/_helpers.py
crusaderky: I thought about it; given the extent of the special-casing needed, I think it is more sensible to directly fix the upstream libraries and eventually completely remove this.

lucascolley: okay — could we add a TODO comment either way just to keep track in the source?

tests/test_funcs.py
crusaderky: https://github.com/pydata/sparse/issues/855

PR #182, opened 2025-03-25
DEV: Configure pixi for Intel Macs

lithomas1:
Without this, if you're on an Intel Mac like me, the tasks won't work properly (in fact, no tasks will show up).
Suprisingly, there is no warning.

To test this, I guess we can add macos-13 testing in Github Actions.
It's probably not be worth it, though (since this is a pure Python package anyways).

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -31,6 +31,24 @@ environments:
       - conda: https://prefix.dev/conda-forge/linux-64/tk-8.6.13-noxft_h4845f30_101.conda
       - conda: https://prefix.dev/conda-forge/noarch/tzdata-2025b-h78e105d_0.conda
       - pypi: .
+      osx-64:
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
ah yes that was an oversight, thanks!

lucascolley:
> To test this, I guess we can add macos-13 testing in Github Actions.
It's probably not be worth it, though (since this is a pure Python package anyways).

Yeah, let's leave that for now.

jakirkham:
> Without this, if you're on an Intel Mac like me, the tasks won't work properly (in fact, no tasks will show up).
> Suprisingly, there is no warning.

Thomas did you already file a Pixi issue about this behavior?

lithomas1:
Good idea, I'll put in a report later this week.

lithomas1:
Submitted https://github.com/prefix-dev/pixi/issues/3441

PR #184, opened 2025-03-26
Add expit function

OmarManzoor:
CC: @lucascolley 

I started with adding this simple function although I am not sure whether we need to add it at all. However does the format look fine. I think we might need more tests that test for dtypes and devices.

Also I am not sure how do I set this up locally to run tests. Do I need to install it? I was simply trying to configure the tests in Visual Studio code on a windows machine but it gives the following error:

```
from array_api_extra._lib import Backend
E   ModuleNotFoundError: No module named 'array_api_extra'
```

### DIFF
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -913,3 +913,29 @@ def sinc(x: Array, /, *, xp: ModuleType | None = None) -> Array:
         xp.asarray(xp.finfo(x.dtype).eps, dtype=x.dtype, device=_compat.device(x)),
     )
     return xp.sin(y) / y
+
+
+def expit(x: Array, /, *, xp: ModuleType | None = None) -> Array:
+    """
+    Return the expit function.
+
+    The expit function, also known as the logistic sigmoid function.
+    It is the inverse of the logit function.
+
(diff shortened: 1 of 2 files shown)


lucascolley:
> Also I am not sure how do I set this up locally to run tests. Do I need to install it?

thanks @OmarManzoor, have you read http://data-apis.org/array-api-extra/contributing.html ? That details the recommended development workflow.

If you don't want to use Pixi, then you could probably fix the error you are seeing with `pip install -e .`.

> I am not sure whether we need to add it at all

I think we could. But I think I would actually prefer that this was first implemented as a "generic_implementation" in https://github.com/scipy/scipy/blob/main/scipy/special/_support_alternative_backends.py. After that point, we could consider making it available via xpx for older SciPy versions. Would you be willing to open a PR at SciPy for that?

OmarManzoor:
@lucascolley Thank you for your reply. Currently I think it might be a bit difficult for me to also setup and contribute on scipy. 

Also thanks for sharing the contributing guide. I missed reading that.

In any case I think let's close this PR for now.

lucascolley:
Sure, we can close this. Would you like to try another function? Maybe `isin`?

PR #186, opened 2025-03-26
DEV: pin CUDA variant for PyTorch

crusaderky:


pyproject.toml
```diff
@@ -154,24 +155,22 @@ jax = "*"
...
 
 # Backends that require a GPU host and a CUDA driver
+# Note: JAX and PyTorch automatically install CUDA variants
+# thanks to the `system-requirements` below.
 [tool.pixi.feature.cuda-backends]
 system-requirements = { cuda = "12" }
 
 [tool.pixi.feature.cuda-backends.target.linux-64.dependencies]
 cupy = "*"
-jaxlib = { version = "*", build = "cuda12*" }
```
lucascolley, line 162: @traversaro could I double check that this is correct?

crusaderky: > for example as cuda 12 was dropped and only cuda 13 is supported [...] the solver will just silently install a cpu-only version of jaxlib

This is a strong argument against this PR. I'll make it explicit again.

crusaderky: @lucascolley ready for review again

lucascolley: thanks for the fast help as always @traversaro !! much appreciated

lucascolley (approved):
thanks Guido

PR #265, opened 2025-04-15
MAINT: Don't use `Backend` for delegation

crusaderky:
Follow-up from https://github.com/data-apis/array-api-extra/pull/221#discussion_r2033668108

Do not use the `Backend` enum outside of testing.

This enum contains a lot of "backends" that are just variant duplicates, which is an artifact of using this enum to parametrize the `xp` fixture. Which is a decent hack for pytest, but which doesn't make any sense in the context of the dispatch system.

This also allows for quite a bit of simplification without adding any weight to the delegate module itself.

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -4,31 +4,20 @@ from collections.abc import Sequence
 from types import ModuleType
 from typing import Literal
 
-from ._lib import Backend, _funcs
-from ._lib._utils._compat import array_namespace
+from ._lib import _funcs
+from ._lib._utils._compat import (
+    array_namespace,
+    is_cupy_namespace,
+    is_dask_namespace,
+    is_jax_namespace,
+    is_numpy_namespace,
+    is_pydata_sparse_namespace,
+    is_torch_namespace,
+)
 from ._lib._utils._helpers import asarrays
 from ._lib._utils._typing import Array
 
 __all__ = ["isclose", "pad"]
-
-
-def _delegate(xp: ModuleType, *backends: Backend) -> bool:
-    """
-    Check whether `xp` is one of the `backends` to delegate to.
-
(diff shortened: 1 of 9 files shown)


src/array_api_extra/_lib/_backends.py
```diff
@@ -1,58 +1,34 @@
-"""Backends with which array-api-extra interacts in delegation and testing."""
+"""Backends with which array-api-extra runs its tests against."""
```
lucascolley, line 1:
```suggestion
"""Backends against which array-api-extra runs its tests."""
```

src/array_api_extra/_delegation.py
```diff
@@ -159,14 +153,19 @@ def pad(
...
+    if is_torch_namespace(xp):
         pad_width = xp.asarray(pad_width)
         pad_width = xp.broadcast_to(pad_width, (x.ndim, 2))
         pad_width = xp.flip(pad_width, axis=(0,)).flatten()
         return xp.nn.functional.pad(x, tuple(pad_width), value=constant_values)  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]
 
-    if _delegate(xp, Backend.NUMPY, Backend.JAX, Backend.CUPY, Backend.SPARSE):
-        return xp.pad(x, pad_width, mode, constant_values=constant_values)
-
     return _funcs.pad(x, pad_width, constant_values=constant_values, xp=xp)
```
lucascolley, line 171: why remove this delegation?

crusaderky: I didn't, I just moved it two lines above for the sake of coherency with `isclose`

lucascolley: ah, sorry, I see that now

lucascolley (approved):

PR #269, opened 2025-04-21
TST: `xfail_xp_backend(strict=False)`

crusaderky:
- Add `strict=False` optional parameter to `xfail_xp_backend`
- raise on spurious kwargs

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -27,7 +27,7 @@ from .conftest import np_compat
 lazy_xp_function(in1d, jax_jit=False, static_argnames=("assume_unique", "invert", "xp"))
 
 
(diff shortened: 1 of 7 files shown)


tests/test_helpers.py
```diff
@@ -27,7 +27,7 @@
 lazy_xp_function(in1d, jax_jit=False, static_argnames=("assume_unique", "invert", "xp"))
 
 
-@pytest.mark.xfail_xp_backend(Backend.SPARSE, reason="no unique_inverse")
+@pytest.mark.skip_xp_backend(Backend.SPARSE, reason="no unique_inverse")
```
crusaderky, line 30: This is an optional function in the Standard

lucascolley (approved):
thanks Guido, LGTM!

PR #283, opened 2025-04-25
ENH: `allow_dask_compute=True` instead of 999

crusaderky:
scipy developers get confused by
`lazy_xp_function(func, allow_dask_compute=999)`

Allow writing
`lazy_xp_function(func, allow_dask_compute=True)`
instead.


lucascolley (approved):
LGTM, thanks!

Issue #270, opened 2025-04-21
ENH: `lazy_xp_function` support for wrapped return values


crusaderky:
`scipy.stats.ttest_ind` fails when wrapped with `lazy_xp_function`.
The reason is that the function returns a custom subclass of NamedTuple.
`@jax.jit` automatically repacks returned lists, tuples, and NamedTuples of arrays, but fails with custom classes.
This however is an artefact specific of `lazy_xp_function`; in real life, end users will unpack and consume the return value of `ttest_ind` within the scope of the jit.

In other words, this fails:

```python
from scipy.stats import ttest_ind
xpx.lazy_xp_function(ttest_ind)

def test1(xp):
    x = xp.asarray([.1, .2])
    y = xp.asarray([.3])
    res = ttest_ind(x, y)
    # res = TtestResult(statistic=np.float64(-1.7320508075688765), pvalue=np.float64(0.3333333333333334), df=np.float64(1.0))

FAILED test1.py::test1[jax.numpy] - TypeError: TtestResult.__init__() missing 4 required positional arguments: 'df', 'alternative', 'standard_error', and 'estimate'
```
as it is equivalent to
```python
>>> from scipy.stats import ttest_ind
>>> jitted = jax.jit(ttest_ind)
>>> jitted(jnp.asarray([.1, .2]), jnp.asarray([.3])
TypeError: TtestResult.__init__() missing 4 required positional arguments: 'df', 'alternative', 'standard_error', and 'estimate'
```

However, in real-life users will not write the above, but will write instead something like
```python
>>> from scipy.stats import ttest_ind
>>> @jax.jit
>>> def f(x, y):
...     res = ttest_ind(x, y)
...     # Stand-in for some post-processing
...     return res.statistic, res.pvalue, res.df)
>>> f(jnp.asarray([.1, .2]), jnp.asarray([.3])
(Array(-1.7320508, dtype=float32),
 Array(0.3333333, dtype=float32),
 Array(1., dtype=float32))
```

# Proposed design
Use pickle hooks to automatically unpack and repack complex return values.

PR #284, opened 2025-04-25
ENH: `jax_autojit`

crusaderky:
Supercharge `xpx.testing.lazy_xp_function`.

This is a fairly high level change; @rgommers @pearu @jakevdp I could use your feedback.

Matching SciPy PR: https://github.com/scipy/scipy/pull/22909

# Automatic static arguments
scipy has seen a proliferation of `lazy_xp_function(func, static_argnames=(...))` in the initial section of many test modules. This information is only useful for JAX and is quite verbose.
With https://github.com/scipy/scipy/pull/22686, this informational quirk that is both specific to JAX and to unit tests moves to the implementation modules.

With this PR, `lazy_xp_function` no longer accepts parameters `static_argnames` and `static_argnums`. Instead, all arguments that are not JAX arrays are automatically treated as static. Note that this behaviour is generally desirable in unit testing but a bad idea in production. Consider:

```python
def f(x: Array, y: float, plus: bool) -> Array:
    return x + y if plus else x - y

j1 = jax.jit(f, static_argnames="plus")
j2 = jax_autojit(f)
```
`jax_autojit` is a new _internal_ function of array-api-extra.
In the above example, `j2` requires a lot less setup to be tested effectively, but on the flip side it means that it will be re-traced for every different value of y, which likely makes it not fit for purpose in production.

To clarify: `jax_autojit` is applied and removed on the fly within tests with the `xp` fixture and it is never used outside of unit testing. 

# Wrapped inputs and output
There are a few cases of scipy functions returning bespoke containers with arrays inside instead of simple tuples, namedtuples, lists, or dicts of arrays.

Two such examples are
- scipy.stats.ttest_ind
- scipy.spatial.Rotation and all functions and methods that return it (discussion: https://github.com/scipy/scipy/pull/22777#discussion_r2057751657)

In main, these can't be tested with `lazy_xp_function`, and as of https://github.com/scipy/scipy/pull/22686 the issue also impacts documentation.

This PR lifts this restriction and allows completely arbitrary objects as parameters and as return values of the functions. If these objects internally contain JAX arrays, `lazy_xp_function` will now automatically extract them, pass them through the JIT, and reassemble everything for the test function to observe the result.   

The rationale is that, in real life, users are unlikely to wrap the scipy functions with jax.jit directly; instead they are more likely to consume their outputs in their own functions _and then_ wrap those with jax.jit:
```python
@jax.jit
def my_user_function(x):
    y = scipy.stats.ttest_ind(x)
    return my_user_consume(y)  # returns array or tuple of arrays
```

# Static non-hashable arguments
Non-hashable objects can now be static.
Consider:
```python
def f(x: Array, verbs: Sequence[str]) -> Array:
    ...
    
lazy_xp_function(f, static_argnames="verbs")

def test_f(xp):
    x = ...
    y = f(x, ("foo", "bar", "baz"))  # OK
    y = f(x, ["foo", "bar", "baz"])  # Fails
```
This PR fixes it; all objects in the input and output of the function need only be hashable _or_ pickleable.

# Dask materialization raises inside a container

This also fixes a Dask-specific bug where the graph materialization raises:
```python
@pytest.mark.skip_xp_backends("jax.numpy", reason="raise inside jax.pure_callback")
def test1(xp):
    with pytest.raises(Exception):
        f(x)
```
The above works when f returns plain Dask array objects or tuples or lists thereof, even if the graph would otherwise be discarded and never computed, thanks to the `lazy_xp_function` machinery, but used to fail when the return value is an opaque container with Dask arrays inside. This PR fixes it.


src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -306,3 +331,233 @@ def capabilities(xp: ModuleType) -> dict[str, int]:
...
+    instances: list[T] = []
+    rest: list[object] = []
+
+    class Pickler(pickle.Pickler):  # numpydoc ignore=GL08
+        """
+        Use the `pickle.Pickler.persistent_id` hook to extractt objects.
+        """
+
+        @override
+        def persistent_id(self, obj: object) -> Literal[0, 1, None]:  # pyright: ignore[reportIncompatibleMethodOverride]  # numpydoc ignore=GL08
```
crusaderky, line 415: To say that I am _aggrieved_ by pyright and numpydoc would be a gentle understatement at this point.

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -306,3 +331,233 @@ def capabilities(xp: ModuleType) -> dict[str, int]:
...
+
+    >>> pickle_unflatten(["foo", "bar", "baz"], rest)
+    {1: "foo", 2: ["bar", <NS>, "baz"]}
+    """
+    instances: list[T] = []
+    rest: list[object] = []
+
+    class Pickler(pickle.Pickler):  # numpydoc ignore=GL08
+        """
+        Use the `pickle.Pickler.persistent_id` hook to extractt objects.
```
crusaderky, line 411:
```suggestion
        Use the `pickle.Pickler.persistent_id` hook to extract objects.
```

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -306,3 +331,233 @@ def capabilities(xp: ModuleType) -> dict[str, int]:
...
+    def __init__(self, obj: T) -> None:  # numpydoc ignore=GL08
+        self._register()
+        self.obj = obj
+
+    @classmethod
+    def _register(cls):  # numpydoc ignore=SS06
+        """
+        Register upon first use instead of at import time, to avoid
+        globally importing JAX.
+        """
```
crusaderky, line 512: Aside note on design: Dask avoids this exact problem by not requiring any decorator, and using unique dunder methods called `__dask_<...>__`

src/array_api_extra/_lib/_utils/_helpers.py
crusaderky, line 1: Note: everything in this module is a private helper.

src/array_api_extra/testing.py
```diff
@@ -36,13 +39,22 @@ def override(func: object) -> object:
 _ufuncs_tags: dict[object, dict[str, Any]] = {}  # type: ignore[explicit-any]
 
 
+class Deprecated(enum.Enum):
+    """Unique type for deprecated parameters."""
+
+    DEPRECATED = 1
+
+
+DEPRECATED = Deprecated.DEPRECATED
```
crusaderky, line 48: I couldn't find anything to help here in the standard library, and I find that odd. Did I miss anything? I was expecting something like `@functools.deprecated_args("static_argnames", "static_argnums")`.

pearu (changes requested):
I have a minor nit and a note, otherwise it looks good to me as it reduces the maintenance burden of adding/updating `static_argnames/argnums` arguments.

Btw, I enjoyed reading `pickle_flatten` and `pickle_flatten` implementations as it felt like less is more.

Thanks, @crusaderky!

jakevdp:
For what it's worth, one of the main reasons we haven't implemented something like this in JAX is because it tends to negatively impact dispatch times. The output of `jax.jit` is a C++ level callable, that directly dispatches to the compiled kernel. This is important because if you put a layer of Python logic in the dispatch path to inspect all the arguments and determine whether to treat them as static or not (and possibly re-compile based on this information) then it severely impacts dispatch times.

With that background, I'd like a bit of clarification here: is this intended for use only in testing paths, or do you imagine this should be used within the dispatch path for libraries like scipy that implement the Array API?

lucascolley:
I think this section of the top post answers your question, Jake:

> To clarify: `jax_autojit` is applied and removed on the fly within tests with the `xp` fixture and it is never used outside of unit testing.

src/array_api_extra/testing.py
crusaderky: Good pointer; however it requires Python >=3.13 so I can't use it :disappointed:

crusaderky:
https://github.com/scipy/scipy/pull/22909 does not cause any issues to crop up. 
This PR is ready for final review.

lucascolley:
@crusaderky will this close gh-270?

crusaderky:
> @crusaderky will this close [gh-270](https://github.com/data-apis/array-api-extra/issues/270)?

Yes, it does!

src/array_api_extra/testing.py
steppi: You could get it from [typing-extensions](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated). It looks like `typing-extensions` is already a dependency too.

crusaderky: It isn't. array-api-extra exclusively depends on array-api-compat exclusively, which in turn has no dependencies (not even numpy).

steppi: I see, it's only in the lint dependencies, I skimmed too quickly.

https://github.com/data-apis/array-api-extra/blob/84b1f0b0c0eb11c3b1acd2bb9aa3f70bac429a37/pyproject.toml#L56-L57

lucascolley:
Yes, I took a look through from a non-expert perspective and nothing seems untoward!

rgommers (commented):
This is very nice, it makes testing of JAX more generic and much easier to understand for maintainers and new contributors to SciPy et al. who don't have much (or any) JAX experience. The test simplifications in this PR show big the difference is.

👍🏼 for the PR description as well, that was very useful to me - and I expect others are going to refer to it in the future.

> Btw, I enjoyed reading pickle_flatten and pickle_flatten implementations as it felt like less is more.

💯 agreed

I have some small non-blocking doc suggestions only. Those can wait or be left alone; I'll go ahead and merge this PR now so we call pull it in to unblock the SciPy release.

The machinery added here is quite complex, so it's very well possible I missed something when reviewing. I tested fairly extensively with SciPy and all looks good though, so nothing major should be off here.

Thanks @crusaderky & reviewers!

PR #285, opened 2025-04-25
MAINT: array_api_compat tweaks

crusaderky:
Import `to_device` and tweak type annotations


src/array_api_extra/_lib/_utils/_compat.pyi
```diff
@@ -12,29 +13,33 @@ from ._typing import Array, Device
 
 # pylint: disable=missing-class-docstring,unused-argument
 
-class Namespace(ModuleType):
-    def device(self, x: Array, /) -> Device: ...
```
crusaderky, line 16: this was plain wrong

lucascolley: aha, good catch

lucascolley (approved):

PR #300, opened 2025-05-18
ENH: support PyTorch `device='meta'`

crusaderky:
- Add support for PyTorch `meta` device. Specifically, this PR fixes:
  - `apply_where`
  - `nunique`
  - `_testing.xp_assert_equal`
  - `_testing.xp_assert_close`
  - `_testing.xp_assert_less`
- Test PyTorch input->output device propagation
  - when input device is `meta` but default device is `cpu`;
  - on pixi environments `tests_cuda` and `dev_cuda`, when input device is `cpu` but default device is `cuda`

XREF https://github.com/data-apis/array-api/issues/945

### DIFF
--- a/src/array_api_extra/_lib/_utils/_helpers.py
+++ b/src/array_api_extra/_lib/_utils/_helpers.py
@@ -29,8 +29,9 @@ from ._compat import (
     is_jax_namespace,
     is_numpy_array,
     is_pydata_sparse_namespace,
+    is_torch_namespace,
 )
-from ._typing import Array
+from ._typing import Array, Device
 
 if TYPE_CHECKING:  # pragma: no cover
     # TODO import from typing (requires Python >=3.12 and >=3.13)
@@ -300,7 +301,7 @@ def meta_namespace(
     return array_namespace(*metas)
 
 
-def capabilities(xp: ModuleType) -> dict[str, int]:
+def capabilities(xp: ModuleType, *, device: Device | None = None) -> dict[str, int]:
     """
     Return patched ``xp.__array_namespace_info__().capabilities()``.
 
@@ -311,6 +312,8 @@ def capabilities(xp: ModuleType) -> dict[str, int]:
     ----------
     xp : array_namespace
         The standard-compatible namespace.
+    device : Device, optional
+        The device to use.
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


tests/test_testing.py
crusaderky, line 1: I want to add an explicit test for `xp_assert_equal` etc. but it's blocked by https://github.com/data-apis/array-api-extra/pull/301

src/array_api_extra/_lib/_utils/_helpers.py
```diff
@@ -326,6 +329,13 @@ def capabilities(xp: ModuleType) -> dict[str, int]:
         # Fixed in jax >=0.6.0
         out = out.copy()
         out["boolean indexing"] = False
+    if is_torch_namespace(xp):
+        # FIXME https://github.com/data-apis/array-api/issues/945
+        device = xp.get_default_device() if device is None else xp.device(device)
+        if cast(Any, device).type == "meta":  # type: ignore[explicit-any]
```
lucascolley, line 335: Can we cast to `Device` instead?

lucascolley (approved):
LGTM! Just one comment

src/array_api_extra/_lib/_utils/_helpers.py
crusaderky: No, because Device is (deliberately) an alias of `object`

lucascolley: Is `xp.get_default_device()` not always an `object`? Or is `object` not a valid input to `cast`?

I know that this won't remove the need for the `type: ignore`, I just mean to be slightly more explicit.

tests/test_testing.py
lucascolley: now merged

src/array_api_extra/_lib/_utils/_helpers.py
crusaderky: The output of `xp.get_default_device()` is `Any` because `xp` is `ModuleType`.
The issue is that a `Device` (from `device: Device | None` in the function signature) as far as the Array API is concerned, is completely opaque; the `.type` attribute is PyTorch-specific so the linter rightfully points it out.

crusaderky: I've reworked it not to use cast (it's not any better IMHO)

lucascolley: thanks, that makes sense

PR #306, opened 2025-05-29
ENH: add new function `one_hot`

### LINKED ISSUE #305: ENH: new function `one_hot`
See https://github.com/data-apis/array-api/issues/787

This is a bit like `xpx.create_diagonal` in that it's another kind of [array creation function.](https://data-apis.org/array-api/latest/API_specification/creation_functions.html).

NeilGirdhar:
Fixes #305

Questions:
* What should the default output dtype be?  Currently, it's the default float type on Jax/PyTorch, or else float64.  Should I try to get the default float type on the platform: `xp.empty(()).dtype`?

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -17,6 +17,7 @@
     isclose
     kron
     nunique
+    one_hot
     pad
     setdiff1d
     sinc
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -14,10 +14,11 @@ from ._lib._utils._compat import (
     is_pydata_sparse_namespace,
     is_torch_namespace,
 )
+from ._lib._utils._compat import device as get_device
 from ._lib._utils._helpers import asarrays
-from ._lib._utils._typing import Array
+from ._lib._utils._typing import Array, DType
 
-__all__ = ["isclose", "pad"]
+__all__ = ["isclose", "one_hot", "pad"]
 
 
 def isclose(
@@ -110,6 +111,83 @@ def isclose(
         return xp.isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan)
 
     return _funcs.isclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan, xp=xp)
+
+
+def one_hot(
+    x: Array,
+    /,
+    num_classes: int,
+    *,
+    dtype: DType | None = None,
+    axis: int = -1,
+    xp: ModuleType | None = None,
+) -> Array:
+    """
+    One-hot encode the given indices.
+
+    Each index in the input `x` is encoded as a vector of zeros of length `num_classes`
+    with the element at the given index set to one.
+
+    Parameters
+    ----------
+    x : array
+        An array with integral dtype whose values are between `0` and `num_classes - 1`.
+    num_classes : int
+        Number of classes in the one-hot dimension.
+    dtype : DType, optional
+        The dtype of the return value.  Defaults to the default float dtype (usually
+        float64).
+    axis : int, optional
+        Position in the expanded axes where the new axis is placed. Default: -1.
+    xp : array_namespace, optional
+        The standard-compatible namespace for `x`. Default: infer.
+
+    Returns
+    -------
+    array
+        An array having the same shape as `x` except for a new axis at the position
+        given by `axis` having size `num_classes`.  If `axis` is unspecified, it
+        defaults to -1, which appends a new axis.
+
+        If ``x < 0`` or ``x >= num_classes``, then the result is undefined, may raise
+        an exception, or may even cause a bad state.  `x` is not checked.
+
+    Examples
+    --------
+    >>> import array_api_extra as xpx
+    >>> import array_api_strict as xp
+    >>> xpx.one_hot(xp.asarray([1, 2, 0]), 3)
+    Array([[0., 1., 0.],
+          [0., 0., 1.],
+          [1., 0., 0.]], dtype=array_api_strict.float64)
+    """
+    # Validate inputs.
+    if xp is None:
+        xp = array_namespace(x)
+    if not xp.isdtype(x.dtype, "integral"):
+        msg = "x must have an integral dtype."
+        raise TypeError(msg)
+    if dtype is None:
+        dtype = _funcs.default_dtype(xp, device=get_device(x))
+    # Delegate where possible.
+    if is_jax_namespace(xp):
+        from jax.nn import one_hot as jax_one_hot
+
+        return jax_one_hot(x, num_classes, dtype=dtype, axis=axis)
+    if is_torch_namespace(xp):
+        from torch.nn.functional import one_hot as torch_one_hot
+
+        x = xp.astype(x, xp.int64)  # PyTorch only supports int64 here.
+        try:
+            out = torch_one_hot(x, num_classes)
+        except RuntimeError as e:
+            raise IndexError from e
+    else:
+        out = _funcs.one_hot(x, num_classes, xp=xp)
+    out = xp.astype(out, dtype, copy=False)
+    if axis != -1:
+        out = xp.moveaxis(out, -1, axis)
+    return out
 
 
 def pad(
--- a/src/array_api_extra/_lib/_funcs.py
+++ b/src/array_api_extra/_lib/_funcs.py
@@ -375,6 +375,23 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     return xp.squeeze(c, axis=axes)
 
 
+def one_hot(
+    x: Array,
+    /,
+    num_classes: int,
+    *,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
+    """See docstring in `array_api_extra._delegation.py`."""
+    # TODO: Benchmark whether this is faster on the NumPy backend:
+    # if is_numpy_array(x):
+    #     out = xp.zeros((x.size, num_classes), dtype=dtype)
+    #     out[xp.arange(x.size), xp.reshape(x, (-1,))] = 1
+    #     return xp.reshape(out, (*x.shape, num_classes))
+    range_num_classes = xp.arange(num_classes, dtype=x.dtype, device=_compat.device(x))
+    return x[..., xp.newaxis] == range_num_classes
+
+
 def create_diagonal(
     x: Array, /, *, offset: int = 0, xp: ModuleType | None = None
 ) -> Array:
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -22,6 +22,7 @@ from array_api_extra import (
     isclose,
     kron,
     nunique,
+    one_hot,
     pad,
     setdiff1d,
     sinc,
@@ -45,6 +46,7 @@ lazy_xp_function(create_diagonal)
 lazy_xp_function(expand_dims)
 lazy_xp_function(kron)
 lazy_xp_function(nunique)
+lazy_xp_function(one_hot)
 lazy_xp_function(pad)
 # FIXME calls in1d which calls xp.unique_values without size
 lazy_xp_function(setdiff1d, jax_jit=False)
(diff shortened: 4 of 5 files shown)


NeilGirdhar:
Fixes #305

lucascolley:
RE dtype, I think https://github.com/data-apis/array-api/pull/848 will give us something slightly cleaner down the line.

In SciPy we have been using https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L399 with `force_floating=True`. Taking `xp.empty(()).dtype` is probably fine for now though.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +383,60 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        from torch.nn.functional import one_hot
+
+        x = xp.astype(x, xp.int64)  # PyTorch only supports int64 here.
+        try:
+            out = one_hot(x, num_classes)
+        except RuntimeError as e:
+            raise IndexError from e
+        if dtype is None:
+            dtype = xp.float
+        out = xp.astype(out, dtype)
```
lucascolley, line 422: please could you follow the pattern in https://github.com/data-apis/array-api-extra/blob/main/src/array_api_extra/_delegation.py and split out the delegation layer to that file? Input validation can happen there as well.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +380,25 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    *,
+    x_size: int,
+    dtype: DType,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
+    """See docstring in `array_api_extra._delegation.py`."""
+    out = xp.zeros((x.size, num_classes), dtype=dtype)
+    x_flattened = xp.reshape(x, (-1,))
+    if is_numpy_namespace(xp):
+        return at(out)[xp.arange(x_size), x_flattened].set(1)
```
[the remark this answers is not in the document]
NeilGirdhar: Okay, let me know if you just want me to move the logic over.  I just want to do this the most logical way.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,89 @@ def isclose(
...
+        msg = "x must have an integral dtype."
+        raise TypeError(msg)
+    if dtype is None:
+        dtype = xp.empty(()).dtype  # Default float dtype
+    # Delegate where possible.
+    if is_jax_namespace(xp):
+        assert is_jax_array(x)
+        from jax.nn import one_hot as jax_one_hot
+
+        return jax_one_hot(x, num_classes, dtype=dtype, axis=axis)
```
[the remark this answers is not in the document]
lucascolley: hmm, I hadn't thought about that. I don't think it is a blocker right now, but feel free to leave like this for now if you prefer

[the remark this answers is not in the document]
NeilGirdhar: Okay, I'll leave it.  Also, I think Jax's `xp` is `jax.numpy`, so I don't think `xp.nn` would even work.

lucascolley:
> I'm not 100% sure how those links will help in this case? 

the point is that instead of writing

```python
dtype = xp.empty(()).dtype
x = xp.zeros(..., dtype=dtype)
```

we would have

```python
x = xp.zeros(...)
x = xp.astype(x, 'real floating')
```

lucascolley:
> I guess you mean askind, right?

Nope, https://github.com/data-apis/array-api/pull/848 suggests overloading the `dtype` parameter of `astype` for this. Feel free to comment over there if you disagree!

NeilGirdhar:
> suggests overloading the `dtype` parameter of `astype` for this. Feel free to comment over there if you disagree!

Oh, no worries, I don't have a strong opinion.  Just trying to keep up with all the planned changes 😄

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +380,25 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    for i in range(x_size):
+        out = at(out)[i, int(x_flattened[i])].set(1)
```
lucascolley, line 398: this `int` forces Dask to materialise. If that is the best we can do, then you want to set `allow_dask_compute` as explained at https://data-apis.org/array-api-extra/generated/array_api_extra.testing.lazy_xp_function.html

NeilGirdhar: Can dask index using scalar arrays?  If not, I'll do what you suggest.

crusaderky: > Can dask index using scalar arrays? If not, I'll do what you suggest.

it should support them, yes.

NeilGirdhar:
> > * What should the default output dtype be?  Currently, it's the default float type on Jax/PyTorch, or else float64.  Should I try to get the default float type on the platform: `xp.empty(()).dtype`?
> 
> to me this doesn't make much sense. Why shouldn't it be `bool`?

I'd rather follow what the libraries are doing than force double conversion for delegated code.

In general, the reason it's not bool is because these values often serve as the inputs to machine learning algorithms.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,89 @@ def isclose(
...
+        except RuntimeError as e:
+            raise IndexError from e
+        out = xp.astype(out, dtype)
+    else:
+        out = _funcs.one_hot(x, num_classes, x_size=x_size, dtype=dtype, xp=xp)
+
+    if x.ndim != 1:
+        out = xp.reshape(out, (*x.shape, num_classes))
+    if axis != -1:
+        out = xp.moveaxis(out, -1, axis)
```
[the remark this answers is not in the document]
NeilGirdhar: (I moved the first part out to the funcs layer as you implied I should.)

src/array_api_extra/_lib/_funcs.py
NeilGirdhar: I tried factoring it out, but Dask doesn't support indexed assignment either, so I'm skipping the Dask tests

NeilGirdhar:
@lucascolley Read for your review

src/array_api_extra/_lib/_funcs.py
```diff
@@ -8,15 +8,19 @@
 
 from ._at import at
 from ._utils import _compat, _helpers
-from ._utils._compat import array_namespace, is_dask_namespace, is_jax_array
+from ._utils._compat import (
+    array_namespace,
+    is_dask_namespace,
+    is_jax_array,
```
lucascolley, line 14: Are these imports unused?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +379,36 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    out = xp.zeros((x.size, num_classes), dtype=dtype)
+    x_flattened = xp.reshape(x, (-1,))
+    if supports_fancy_indexing:
+        out = at(out)[xp.arange(x_size), x_flattened].set(1)
+    else:
+        for i in range(x_size):
+            x_i = x_flattened[i]
+            if not supports_array_indexing:
+                x_i = int(x_i)
+            out = at(out)[i, x_i].set(1)
```
lucascolley, line 406: @crusaderky have we encountered patterns like this before?

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,90 @@ def isclose(
...
+    axis : int or tuple of ints, optional
+        Position(s) in the expanded axes where the new axis is placed.
+    xp : array_namespace, optional
+        The standard-compatible namespace for `x`. Default: infer.
+
+    Returns
+    -------
+    array
+        An array having the same shape as `x` except for a new axis at the position
+        given by `axis` having size `num_classes`.
```
lucascolley, line 150: Perhaps this can be slightly reworded to account for the case `axis` is unprovided (or the default can be specified for `axis` above)

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,90 @@ def isclose(
...
+
+        If ``x < 0`` or ``x >= num_classes``, then the result is undefined, may raise
+        an exception, or may even cause a bad state.  `x` is not checked.
+
+    Examples
+    --------
+    >>> xp.one_hot(jnp.array([1, 2, 0]), 3)
```
lucascolley, line 157: Can you add `import array_api_extra as xpx` and change this to `xpx.one_hot`. Also please import array-api-strict as `xp` and use `xp` instead of `jnp`. The output should match array-api-strict's output.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,90 @@ def isclose(
...
+    if is_jax_namespace(xp):
+        assert is_jax_array(x)
+        from jax.nn import one_hot as jax_one_hot
+
+        return jax_one_hot(x, num_classes, dtype=dtype, axis=axis)
+    if is_torch_namespace(xp):
+        assert is_torch_array(x)
+        from torch.nn.functional import one_hot as torch_one_hot
+
+        x = xp.astype(x, xp.int64)  # PyTorch only supports int64 here.
```
lucascolley, line 180: Just to check, is this true always? Does PyTorch not have a concept of setting the default int to `int32`?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +379,36 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    num_classes: int,
+    *,
+    supports_fancy_indexing: bool = False,
+    supports_array_indexing: bool = False,
+    dtype: DType,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
+    """See docstring in `array_api_extra._delegation.py`."""
+    x_size = x.size
+    if x_size is None:  # pragma: no cover
```
lucascolley, line 394: Optional nit: could specify the problem (I assume undefined attribute or some such error)

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +379,36 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        msg = "x must have a concrete size."
```
lucascolley, line 395: I think we should specify this in the docstring.

tests/test_funcs.py
```diff
@@ -448,6 +450,78 @@ def test_xp(self, xp: ModuleType):
         )
 
 
+@pytest.mark.skip_xp_backend(
+    Backend.SPARSE, reason="read-only backend without .at support"
+)
+@pytest.mark.skip_xp_backend(
+    Backend.DASK, reason="backend does not yet support indexed assignment"
+)
```
lucascolley, line 458: Is this consistent with our skips elsewhere for the same reason @crusaderky ? (I'm on mobile, not super easy to check)

src/array_api_extra/_delegation.py
NeilGirdhar: Done.

src/array_api_extra/_delegation.py
NeilGirdhar: Done.

src/array_api_extra/_lib/_funcs.py
NeilGirdhar: Good idea, done.

src/array_api_extra/_lib/_funcs.py
NeilGirdhar: Good idea, done.

src/array_api_extra/_delegation.py
NeilGirdhar: If I remove this check, I get:
```
      14 failed
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-int8]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-int16]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-int32]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-uint8]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-uint16]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-uint32]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch-uint64]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-int8]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-int16]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-int32]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-uint8]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-uint16]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-uint32]
         - tests/test_funcs.py:494 TestOneHot.test_int_types[torch:gpu-uint64]
```
which is why I added that test.  PyTorch raises with the message "RuntimeError: one_hot is only applicable to index tensor of type LongTensor. "

lucascolley: right, this lines up with the PyTorch docs.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +114,93 @@ def isclose(
...
+
+    Returns
+    -------
+    array
+        An array having the same shape as `x` except for a new axis at the position
+        given by `axis` having size `num_classes`.  If `axis` is unspecified, it
+        defaults to -1, which appends a new axis.
+
+        If ``x < 0`` or ``x >= num_classes``, then the result is undefined, may raise
+        an exception, or may even cause a bad state.  `x` is not checked.
```
lucascolley, line 154: did you consider checking `x` and always raising an error here? Does JAX always return a zero array?

NeilGirdhar: I considered it, but I don't want to add any steps that might make this function any slower.

In the most common case, the values are not out of bounds.  They're typically something [like this](https://github.com/NeilGirdhar/efax/blob/1c0867457631ef22a10009314a1461468fa013ed/efax/_src/distributions/multinomial.py#L77): converted samples that are guaranteed to be in bounds.

src/array_api_extra/_lib/_funcs.py
crusaderky: No, we didn't.

crusaderky:
@NeilGirdhar not sure you noticed my suggestion above?

> What about `x[..., None] == xp.arange(num_classes, dtype=x.dtype, device=_compact.device(x))`? 

that should be all you need to do to implement `one_hot` in a way that is both performant and Array API compliant.
(the numpy special case _may_ be faster; needs benchmarking).

src/array_api_extra/_lib/_funcs.py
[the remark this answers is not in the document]
NeilGirdhar: Great idea, done.

tests/test_funcs.py
```diff
@@ -448,6 +450,75 @@ def test_xp(self, xp: ModuleType):
         )
 
 
+@pytest.mark.skip_xp_backend(
+    Backend.SPARSE, reason="read-only backend without .at support"
+)
```
crusaderky, line 455:
```suggestion
```
this should now work.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +115,87 @@ def isclose(
...
+    """
+    One-hot encode the given indices.
+
+    Each index in the input `x` is encoded as a vector of zeros of length `num_classes`
+    with the element at the given index set to one.
+
+    Parameters
+    ----------
+    x : array
+        An array with integral dtype and concrete size (``x.size`` cannot be `None`).
```
crusaderky, line 136: I think with the new algorithm none sizes are supported (needs a test though)

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +115,87 @@ def isclose(
...
+        from torch.nn.functional import one_hot as torch_one_hot
+
+        x = xp.astype(x, xp.int64)  # PyTorch only supports int64 here.
+        try:
+            out = torch_one_hot(x, num_classes)
+        except RuntimeError as e:
+            raise IndexError from e
+    else:
+        out = _funcs.one_hot(x, num_classes, xp=xp)
+    out = xp.astype(out, dtype)
```
crusaderky, line 193:
```suggestion
    out = xp.astype(out, dtype, copy=False)
```

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +115,87 @@ def isclose(
...
+            "real floating"
+        ]
+    # Delegate where possible.
+    if is_jax_namespace(xp):
+        assert is_jax_array(x)
+        from jax.nn import one_hot as jax_one_hot
+
+        return jax_one_hot(x, num_classes, dtype=dtype, axis=axis)
+    if is_torch_namespace(xp):
+        assert is_torch_array(x)
```
crusaderky, line 183:
```suggestion
```

tests/test_funcs.py
```diff
@@ -448,6 +450,75 @@ def test_xp(self, xp: ModuleType):
...
+        actual = one_hot(xp.asarray([1, 2, 0]), 3, axis=0)
+        xp_assert_equal(actual, expected)
+
+        actual = one_hot(xp.asarray([1, 2, 0]), 3, axis=-2)
+        xp_assert_equal(actual, expected)
+
+    def test_non_integer(self, xp: ModuleType):
+        with pytest.raises(TypeError):
+            _ = one_hot(xp.asarray([1.0]), 3)
+
```
crusaderky, line 520: please add `test_device` (see many examples elsewhere)

tests/test_funcs.py
NeilGirdhar: It doesn't have arange, but I updated the `reason`.

tests/test_funcs.py
NeilGirdhar: Done

crusaderky (changes requested):
This is almost ready to go; a few nits below.

It's worth noting that the current implementation makes it impossible to write
```python
symbols, idx = xp.unique_inverse(x)
xpx.one_hot(idx, symbols.size)
```
as a pattern to build a one-hot map of arbitrary symbols on Dask, unless you know in advance the maximum number of unique symbols.
This _could_ be fixed in a follow-up but I'm unsure about real-life interest in it.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -375,6 +375,23 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    num_classes: int,
+    *,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
+    """See docstring in `array_api_extra._delegation.py`."""
+    # TODO: Benchmark whether this is faster on the NumPy backend:
+    # if is_numpy_array(x):
+    #     out = xp.zeros((x.size, num_classes), dtype=dtype)
+    #     out[xp.arange(x.size), xp.reshape(x, (-1,))] = 1
+    #     return xp.reshape(out, (*x.shape, num_classes))
```
lucascolley, line 390: If we did add this, I would rather it lived in `_delegation`, as long as the condition remained `is_numpy_array`.

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +115,87 @@ def isclose(
...
+    # Validate inputs.
+    if xp is None:
+        xp = array_namespace(x)
+    if not xp.isdtype(x.dtype, "integral"):
+        msg = "x must have an integral dtype."
+        raise TypeError(msg)
+    if dtype is None:
+        dtype = xp.__array_namespace_info__().default_dtypes(device=get_device(x))[
+            "real floating"
+        ]
```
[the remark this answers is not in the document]
crusaderky: Yes, `xpx.default_float(xp, *, device=None)` would be good.

docs/api-reference.md
```diff
@@ -15,6 +15,7 @@
     expand_dims
     isclose
     kron
+    one_hot
     nunique
```
crusaderky, line 19:
```suggestion
    nunique
    one_hot
```

src/array_api_extra/_delegation.py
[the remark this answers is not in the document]
crusaderky: https://github.com/data-apis/array-api-extra/pull/310

tests/test_funcs.py
```diff
@@ -448,6 +450,78 @@ def test_xp(self, xp: ModuleType):
...
+        )
+        xp_assert_equal(actual, expected)
+
+    def test_axis(self, xp: ModuleType):
+        expected = xp.asarray([[0.0, 1.0, 0.0], [0.0, 0.0, 1.0], [1.0, 0.0, 0.0]]).T
+        actual = one_hot(xp.asarray([1, 2, 0]), 3, axis=0)
+        xp_assert_equal(actual, expected)
+
+        actual = one_hot(xp.asarray([1, 2, 0]), 3, axis=-2)
+        xp_assert_equal(actual, expected)
```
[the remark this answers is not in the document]
NeilGirdhar: Done

tests/test_funcs.py
```diff
@@ -448,6 +450,78 @@ def test_xp(self, xp: ModuleType):
...
+
+    def test_non_integer(self, xp: ModuleType):
+        with pytest.raises(TypeError):
+            _ = one_hot(xp.asarray([1.0]), 3)
+
+    def test_device(self, xp: ModuleType, device: Device):
+        x = xp.asarray([0, 1, 2], device=device)
+        y = one_hot(x, 3)
+        assert get_device(y) == device
```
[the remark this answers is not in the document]
NeilGirdhar: Done

src/array_api_extra/_delegation.py
```diff
@@ -112,6 +113,85 @@ def isclose(
...
+        given by `axis` having size `num_classes`.  If `axis` is unspecified, it
+        defaults to -1, which appends a new axis.
+
+        If ``x < 0`` or ``x >= num_classes``, then the result is undefined, may raise
+        an exception, or may even cause a bad state.  `x` is not checked.
+
+    Examples
+    --------
+    >>> import array_api_extra as xpx
+    >>> import array-api-strict as xp
```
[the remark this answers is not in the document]
NeilGirdhar: Done

NeilGirdhar:
Thanks @lucascolley and @crusaderky for the thorough and quick reviews!

lucascolley:
looks like there was a rebase hiccup

crusaderky (approved):

lucascolley:
thanks again!

PR #318, opened 2025-06-04
MAINT: bump to sparse >=0.17

crusaderky:
Outstanding deficiencies, as far as array-api-extra is concerned:

- Missing `arange`
- Missing `argsort`
- Missing `unique_inverse`
- Functions that return a scalar return np.generic, which belongs to a different namespace
- No `.at` method or equivalent
- https://github.com/pydata/sparse/issues/855
- https://github.com/pydata/sparse/issues/876

CC @hameerabbasi


hameerabbasi:
Hey! Thanks for the list of issues; we'll work through them.

PR #326, opened 2025-06-05
ENH: add `Generator`

NeilGirdhar:


NeilGirdhar:
Note, if Jax's traced decorators were standardized too, it would be possible to pass `JaxGenerator` to them seamlessly.  I imagine it would be similar if Pytorch's traced decorators and its generator.

Issue #194, opened 2025-03-31
DEV: use `pixi` for pre-commit


lucascolley:
discussion at https://github.com/quantity-dev/metrology-apis/pull/15#discussion_r2020715037

example: https://github.com/prefix-dev/pixi/blob/main/.pre-commit-config.yaml

lucascolley:
also https://lefthook.dev/

PR #344, opened 2025-06-30
DEV: replace pre-commit with lefthook+Pixi

lucascolley:
closes gh-194


lucascolley:
```
src/array_api_extra/_lib/_utils/_compat.pyi: error: Duplicate module named "array_api_extra._lib._utils._compat" (also at "src/array_api_extra/_lib/_utils/_compat.py")
src/array_api_extra/_lib/_utils/_compat.pyi: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
src/array_api_extra/_lib/_utils/_compat.pyi: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
```

@jorenham how can I deal with this odd error from mypy?

lucascolley:
fixed by not providing `mypy` with staged files

PR #348, opened 2025-07-02
Remove array-api-strict from 3.10 CI

crusaderky:
This is being discussed here https://github.com/data-apis/array-api-strict/issues/160.
It implements option 2/3 in the linked thread.

### DIFF
--- a/src/array_api_extra/_lib/_backends.py
+++ b/src/array_api_extra/_lib/_backends.py
@@ -58,13 +58,6 @@ class Backend(Enum):  # numpydoc ignore=PR02
         )
 
         marks = []
-        if self.like(Backend.ARRAY_API_STRICT):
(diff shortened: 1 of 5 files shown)


src/array_api_extra/_lib/_backends.py
```diff
@@ -58,13 +58,6 @@ def pytest_param(self) -> Any:
         )
 
         marks = []
-        if self.like(Backend.ARRAY_API_STRICT):
-            marks.append(
-                pytest.mark.skipif(
-                    NUMPY_VERSION < (1, 26),
-                    reason="array_api_strict is untested on NumPy <1.26",
-                )
-            )
```
crusaderky, line 67: numpy 1.26 is the first release with Python 3.12 support

PR #395, opened 2025-08-21
ENH: `at`: add `__setitem__` fancy indexing fallback

amacati:
Fancy indexing is currently not supported for `__setitem__`, which blocks some PRs in scipy (https://github.com/scipy/scipy/pull/23425).

As discussed in https://github.com/data-apis/array-api/issues/864#issuecomment-3167352347, all frameworks of the array api already implement this feature, but not necessarily in a consistent manner for duplicate indices. It is currently not part of the standard. This PR adds a workaround for array api strict to allow fancy indexing in `xpx.at(x, ...)`.

### DIFF
--- a/src/array_api_extra/_lib/_at.py
+++ b/src/array_api_extra/_lib/_at.py
@@ -11,6 +11,8 @@ from typing import TYPE_CHECKING, ClassVar, cast
 from ._utils import _compat
 from ._utils._compat import (
     array_namespace,
+    is_array_api_obj,
+    is_array_api_strict_namespace,
     is_dask_array,
     is_jax_array,
     is_torch_array,
@@ -351,6 +353,33 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         # https://github.com/pytorch/pytorch/issues/150017
         if is_torch_array(y):
             y = xp.astype(y, x.dtype, copy=False)
+
+        # Work around lack of fancy indexing __setitem__ support in array-api-strict.
+        if (
+            is_array_api_strict_namespace(xp)
+            and is_array_api_obj(idx)
+            and xp.isdtype(idx.dtype, "integral")
+            and out_of_place_op is None  # only use for set()
+        ):
+            # Vectorize the operation using boolean indexing
+            # For non-unique indices, take the last occurrence. This requires creating
+            # masks for x and y that create matching shapes.
+            unique_indices, first_occurrence_mask = xp.unique_inverse(idx)
+            x_mask = xp.any(xp.arange(x.shape[0])[..., None] == unique_indices, axis=-1)
+            # Get last occurrence of each unique index
+            cmp = unique_indices[:, None] == unique_indices[None, :]
+            # Ignore later matches
+            lower_tri_mask = (
+                xp.arange(y.shape[0])[:, None] >= xp.arange(y.shape[0])[None, :]
+            )
+            masked_cmp = cmp & lower_tri_mask
+            # For each position i, count how many matches occurred before i
+            prior_matches = xp.sum(xp.astype(masked_cmp, xp.int32), axis=-1)
+            # Last occurrence has highest match count
+            y_mask = prior_matches == xp.max(prior_matches, axis=-1)
+            # Apply the operation only to last occurrences
+            x[x_mask] = y[y_mask]
+            return x
 
         # Backends without boolean indexing (other than JAX) crash here
         if in_place_op:  # add(), subtract(), ...
--- a/src/array_api_extra/_lib/_at.py
+++ b/src/array_api_extra/_lib/_at.py
@@ -150,6 +150,19 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         >>> xpx.at(jnp.asarray([123]), jnp.asarray([0, 0])).add(1)
         Array([125], dtype=int32)
 
+    For frameworks that don't support fancy indexing by default, e.g. array-api-strict,
+    we implement a workaround for 1D integer indices and ``xpx.at().set``. Assignments
+    with multiple occurences of the same index always choose the last occurence. This is
+    consistent with numpy's behaviour, e.g.::
+
+        >>> import numpy as np
+        >>> import array_api_strict as xp
+        >>> import array_api_extra as xpx
+        >>> xpx.at(np.asarray([0]), np.asarray([0, 0])).set(np.asarray([2, 3]))
+        array([3])
+        >>> xpx.at(xp.asarray([0]), xp.asarray([0, 0])).set(xp.asarray([2, 3]))
+        Array([3], dtype=array_api_strict.int64)
+
     See Also
     --------
     jax.numpy.ndarray.at : Equivalent array method in JAX.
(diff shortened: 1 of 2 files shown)


lucascolley (commented):
thanks @amacati !

lucascolley:
lint failures are real but trivial

src/array_api_extra/_lib/_at.py
```diff
@@ -148,6 +150,19 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
         >>> xpx.at(jnp.asarray([123]), jnp.asarray([0, 0])).add(1)
         Array([125], dtype=int32)
 
+    For frameworks that don't support fancy indexing by default, e.g. array-api-strict,
+    we implement a workaround for 1D integer indices and ``xpx.at().set``. Assignments
+    with multiple occurrences of the same index always choose the last occurrence. This
+    is consistent with numpy's behaviour, e.g.::
```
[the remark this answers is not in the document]
crusaderky: You should also clarify that support for integer array indices is exclusively and narrowly for a single index along the first axis, and it must not be represented as a tuple.

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
...
+                    oob_index = int(u_idx[first_oob_idx])
+                elif xp.any(neg_oob := u_idx_pos < 0):
+                    first_oob_idx = xp.argmax(xp.astype(neg_oob, xp.int32))
+                    oob_index = int(u_idx[first_oob_idx])
+                if oob_index is not None:
+                    msg = (
+                        f"index {oob_index} is out of bounds for array of "
+                        f"shape {x.shape}"
+                    )
+                    raise IndexError(msg) from e
```
crusaderky, line 409:
```suggestion
                if (
                    not is_lazy_array(u_idx_pos)
                    and (xp.any(u_idx_pos < 0) or xp.any(u_idx_pos >= x.shape[0]))
                ):
                    msg = (
                        f"index or indices out of bounds for array of "
                        f"shape {x.shape}"
                    )
                    raise IndexError(msg) from e
```
felt overengineered?

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
...
+
+                x_mask = xp.any(xp.arange(x.shape[0])[..., None] == u_idx_pos, axis=-1)
```
crusaderky, line 411:
```suggestion
                x_rng = xp.arange(x.shape[0], device=device(u_idx_pos))
                x_mask = xp.any(x_rng[..., None] == u_idx_pos, axis=-1)
```
Could you add a comment explaning what you're doing here?

crusaderky: I think you need to add in the documentation above a note warning that the implementation is quadratic.

If x is 10 MiB and the u_idx_pos is 10 MiB, _this line transitorily consumes 100 terabytes of RAM._
Have you considered using `searchsorted`?

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
...
+                    x[x_mask] = y
+                    return x
+                if y.shape[0] != idx.shape[0]:
+                    msg = (
+                        f"shape mismatch: value array of shape {y.shape} could not be "
+                        f"broadcast to indexing result of shape {idx.shape}"
+                    )
+                    raise ValueError(msg) from e
+                # If not, create a mask for y. Get last occurrence of each unique index
+                cmp = u_idx_pos[:, None] == u_idx_pos[None, :]
```
crusaderky, line 423:
```suggestion
                cmp = xp.eye(u_idx_pos.shape[0], dtype=xp.bool, device=device(u_idx_pos))
```

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
...
+                # Ignore later matches
+                lower_tri_mask = (
+                    xp.arange(y.shape[0])[:, None] >= xp.arange(y.shape[0])[None, :]
+                )
+                masked_cmp = cmp & lower_tri_mask
```
crusaderky, line 428: we could rewrite these lines as
```python
n = u_idx_pos.shape[0]
cmp = xp.eye(n)
lower_tri_mask = xp.tril(xp.ones((n, n), dtype=xp.bool))
```
masked_cmp is always equal to cmp by construction, unless I'm missing something.

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
...
+                # For each position i, count how many matches occurred before i
+                prior_matches = xp.sum(xp.astype(masked_cmp, xp.int32), axis=-1)
```
crusaderky, line 430: philosophical nit: automatically use native int
```suggestion
                prior_matches = xp.sum(xp.where(masked_cmp, 1, 0), axis=-1)
```

src/array_api_extra/_lib/_at.py
```diff
@@ -355,9 +370,70 @@ def _op(
         # Backends without boolean indexing (other than JAX) crash here
         if in_place_op:  # add(), subtract(), ...
             x[idx] = in_place_op(x[idx], y)
```
crusaderky, line 372: These remain broken.

src/array_api_extra/_lib/_at.py
```diff
@@ -148,6 +150,19 @@ class at:  # pylint: disable=invalid-name  # numpydoc ignore=PR02
...
+    with multiple occurrences of the same index always choose the last occurrence. This
+    is consistent with numpy's behaviour, e.g.::
+
+        >>> import numpy as np
+        >>> import array_api_strict as xp
+        >>> import array_api_extra as xpx
+        >>> xpx.at(np.asarray([0]), np.asarray([0, 0])).set(np.asarray([2, 3]))
+        array([3])
+        >>> xpx.at(xp.asarray([0]), xp.asarray([0, 0])).set(xp.asarray([2, 3]))
+        Array([3], dtype=array_api_strict.int64)
```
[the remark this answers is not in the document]
amacati: The aim is to show that np's and xpx's behavior is identical. For torch tensors on the GPU you would see 
```python
>>> xpx.at(torch.tensor([0]).cuda(), torch.tensor([0, 0]).cuda()).set(torch.tensor([2, 3]).cuda())
torch.Tensor([2], dtype=torch.int64)
```

src/array_api_extra/_lib/_at.py
amacati: Not valid in array-api-strict:

> TypeError: Two scalars not allowed, got type(x1) = <class 'int'> and type(x2) =<class 'int'>

src/array_api_extra/_lib/_at.py
amacati: See above.

lucascolley:
To job my memory, is the blocker for the standard that existing libraries implement orthogonal semantics, so standardising one set of behaviour would mean a significant break for at least some library?

lucascolley:
Could this PR be reworked into a documentation note, explaining what is supported for known existing array libraries (but not arbitrary compliant libraries)

amacati:
Has this not been investigated as part of the effort to standardize integer indexing?

PR #398, opened 2025-08-22
ENH: Implement `nan_to_num` function

### LINKED ISSUE #304: ENH: add `nan_to_num`
Cross-referencing: https://github.com/data-apis/array-api/issues/878

TL;DR: a proposal was made on the Array API repo to add `nan_to_num` to the specification; however, there wasn't much appetite for making such an addition. Instead, array-api-extra seemed like a better choice.

As such, I'm moving discussion for adding this API to this repo.

cc @ogrisel

paddyroddy:
Fixes #304. This is still in progress but want to see what would happen in CI.

`nan_to_num` is a weird function that has unexpected behaviour IMO (e.g. infinities are converted too) but feel, for now, we should keep it as downstream behaviours would expect.

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -126,7 +126,7 @@ def nan_to_num(
     If `x` is inexact, NaN is replaced by zero or by the user defined value in the
     `fill_value` keyword, infinity is replaced by the largest finite floating
     point value representable by ``x.dtype``, and -infinity is replaced by the
-    most negative finite floating point values representable by ``x.dtype``.
+    most negative finite floating point value representable by ``x.dtype``.
 
     For complex dtypes, the above is applied to each of the real and
     imaginary components of `x` separately.


src/array_api_extra/_delegation.py
```diff
@@ -113,6 +113,82 @@ def isclose(
...
+    negative finite floating point values representable by ``x.dtype``.
+
+    For complex dtypes, the above is applied to each of the real and
+    imaginary components of `x` separately.
+
+    If `x` is not inexact, then no replacements are made.
+
+    Parameters
+    ----------
+    x : array, float, complex
```
lucascolley, line 138: I think it is more idiomatic to use `or` instead of a comma for the separator here: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

src/array_api_extra/_lib/_funcs.py
```diff
@@ -738,6 +738,48 @@ def kron(
     return xp.reshape(result, res_shape)
 
 
+def nan_to_num(
+    x: Array,
+    /,
+    *,
+    fill_value: int | float | complex = 0.0,
+    xp: ModuleType | None = None,
```
lucascolley: I mentioned this in person but I think we should make `xp` required instead of optional in the internal `_funcs` implementation.

tests/test_funcs.py
```diff
@@ -941,6 +943,58 @@ def test_xp(self, xp: ModuleType):
...
+                complex(xp.nan, xp.inf),
+            ]
+        )
+        xp_assert_equal(
+            nan_to_num(a),
+            xp.asarray(
+                [
+                    infinity + 0j,
+                    0 + 0j,
+                    0 + 1j * infinity
```
lucascolley:
```suggestion
                    0 + 1j*infinity
```

src/array_api_extra/_delegation.py
```diff
@@ -113,6 +113,82 @@ def isclose(
...
+    *,
+    fill_value: int | float | complex = 0.0,
+    xp: ModuleType | None = None
+) -> Array:
+    """
+    Replace NaN with zero and infinity with large finite numbers (default
+    behaviour).
+
+    If `x` is inexact, NaN is replaced by zero or by the user defined value in
+    `nan` keyword, infinity is replaced by the largest finite floating point
```
lucascolley, line 127: `nan` is now `fill_value`

src/array_api_extra/_delegation.py
```diff
@@ -113,6 +113,83 @@ def isclose(
...
+    values representable by ``x.dtype`` and -infinity is replaced by the most
+    negative finite floating point values representable by ``x.dtype``.
+
+    For complex dtypes, the above is applied to each of the real and
+    imaginary components of `x` separately.
+
+    If `x` is not inexact, then no replacements are made.
```
lucascolley, line 134: but the implementation calls `perform_replacements` for integer dtypes

tests/test_funcs.py
paddyroddy: `ruff` insists on this

src/array_api_extra/_delegation.py
paddyroddy: Good point, that was copied verbatim from NumPy – but it's not correct

lucascolley (approved):
thanks @paddyroddy, this looks great!

PR #420, opened 2025-09-08
MAINT: renovate to skip JAX

crusaderky:
jax 0.7.0 and later is incompatible with Python 3.10.
jax.0.6.2 and later segfaults on CUDA. Note that CUDA is not in CI.

I've never tampered with renovate before so I'm not 100% sure about what I'm doing.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -29,7 +29,7 @@
     "description": "Block PRs for updates blocked on dropping Python 3.10.",
     "matchManagers": ["pixi"],
     "matchUpdateTypes": ["major", "minor"],
-    "matchPackageNames": ["numpy", "sphinx", "ipython", "sphinx-autodoc-typehints"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -29,7 +29,7 @@
     "description": "Block PRs for updates blocked on dropping Python 3.10.",
     "matchManagers": ["pixi"],
     "matchUpdateTypes": ["major", "minor"],
-    "matchPackageNames": ["numpy", "sphinx", "ipython", "sphinx-autodoc-typehints"],
+    "matchPackageNames": ["numpy", "jax", "jaxlib",  "sphinx", "ipython", "sphinx-autodoc-typehints"],
```
crusaderky, line 32:
```suggestion
    "matchPackageNames": ["numpy", "jax", "jaxlib", "sphinx", "ipython", "sphinx-autodoc-typehints"],
```

lucascolley (approved):
ah yes, that looks right to me, thanks for tackling this Guido!

PR #445, opened 2025-09-26
MAINT: update lock file to unblock CI

lucascolley:


lucascolley:
https://github.com/data-apis/array-api-extra/blob/c94a47fc1f01938db99d07b14d6c4fdd63ca6988/tests/test_funcs.py#L226 and https://github.com/data-apis/array-api-extra/blob/c94a47fc1f01938db99d07b14d6c4fdd63ca6988/tests/test_funcs.py#L242

jorenham:
> @jorenham any advice for the NumPy dtype typing failures that have appeared upon upgrading mypy?

You're comparing a `np.dtype` instance using `is` with a `type[np.float32]`, which indeed is always false. You probably want to instead do `dtype.type is np.float32`

jorenham:
> > > @jorenham any advice for the NumPy dtype typing failures that have appeared upon upgrading mypy?
> > 
> > 
> > You're comparing a `np.dtype` instance using `is` with a `type[np.float32]`, which indeed is always false. You probably want to instead do `dtype.type is np.float32`
> 
> So the problem is that we're annotating `np.float32` with `np.dtype[Any]`? How are we supposed to type that?

is `dtype` an instance of `np.dtype`, e.g. `dtype=np.dtype("float32")'`, or is is a scalar type like `dtype=np.float32`?

lucascolley:
https://github.com/data-apis/array-api-extra/blob/main/tests/test_funcs.py#L209

jorenham:
> A scalar type.

Ah ok, then 

https://github.com/data-apis/array-api-extra/blob/c94a47fc1f01938db99d07b14d6c4fdd63ca6988/tests/test_funcs.py#L217

should instead be `dtype: type[np.generic]` or `dtype: type[np.floating]` or something like that

jorenham:
BTW, hypothesis also has a dedicated strategy for this: https://hypothesis.readthedocs.io/en/latest/reference/strategies.html#hypothesis.extra.numpy.floating_dtypes

PR #451, opened 2025-10-02
ENH: `cov` delegation

adriagarp:
Adds delegation for the `cov` function.

All backends (except PyData Sparse) have a `cov` implementation that accepts a single positional argument.


lucascolley (commented):
thanks @adriagarp !

It looks like the tests are reporting some changes in the return dtypes. Could you write a summary about what changes here?

We may want to:
- wrap the output from `xp.cov` to change dtypes
- make a breaking change to match the behaviour of the underlying libraries

Either way, it would be nice to add return dtype information to the docstring, once we have decided what to do?

adriagarp:
Sure! The test failures are as follows:

- `test_empty` (`jax` and `torch`): for an empty tensor in `jax`, the actual result is `array(-0.)`, while the expected value is `array(nan)`. In `torch`, the test fails because of a `UserWarning`
- `test_combination`and `test_xp` (`torch`): torch defaults to `float32`, but the tests explicitly check for `float64` results
- `test_device` (`torch`): I'm quite lost with this error, since I've never worked with PyTorch. It's as follows:
`NotImplementedError: aten::equal: attempted to run this operator with Meta tensors, but there was no fake impl or Meta kernel registered. You may have run into this message while using an operator with PT2 compilation APIs (torch.compile/torch.export); in order to use this operator with those APIs you'll need to add a fake impl. Please see the following for next steps:  https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html`
Have you seen anything like this before? Any hints? I tried reading the provided link but

For the `test_empty` case, I would propose to check for empty arrays and return `arran(nan)` before calling `cov`, and document this in the docstring. What do you think?

For the `float32` vs `float64`, what is the common approach here? Should we keep backend's behavior, or try to have an unified return dtype?

adriagarp:
@lucascolley I've added explicit dtypes for all tests in `TestCov`, and created the two mentioned issues.

Let me know if anything's missing!

Issue #100, opened 2025-01-18
ENH: delegation tracker


lucascolley:
Where native implementations (more simple than the implementations in `_funcs.py`) are possible for known array library backends, we should add delegation to those implementations, instead of using the array-agnostic implementation. This is for performance reasons, and it may also allow us to overcome limitations in the "agnostic" implementations if they are not fully agnostic.

See https://github.com/data-apis/array-api-extra/blob/main/src/array_api_extra/_delegation.py (TODO: write contributor docs) for how to do this.

- [ ] `atleast_nd`
- [ ] `cov`
- [ ] `create_diagonal`
- [ ] `expand_dims`
- [ ] `kron`
- [ ] `nunique`
- [ ] `pad`
- [ ] `setdiff1d`
- [ ] `sinc`

Cyril-36:
Hi, I’d like to work on adding delegation for broadcast_shapes as part of this tracker. I’ve read the contributing section on how to add delegation to a function and will follow that approach. Please let me know if anyone is already working on broadcast_shapes; if not, I’ll start implementing it and open a PR.

lucascolley:
@Cyril-36 go for it!

lucascolley:
Every item that was on the list here has been checked off. Remaining work:
- check we haven't missed anything
- check that we're happy that the coverage for functions we've checked off is up-to-date

At that point, I think we can close this and just add delegation on an as-requested basis.

lucascolley:
looks like `angle` is still remaining

PR #454, opened 2025-10-02
ENH: `atleast_nd` delegation

Enderdead:
This merge request introduces the `atleast_nd` delegate function (#100).

Unlike other delegate functions, `atleast_nd` does not exist for all major backends. Instead, the functions `atleast_1d`, `atleast_2d`, and `atleast_3d` are available.

In this implementation, I use `getattr` to delegate when ndim is between 1 and 3. I’ve also added tests with ndim=3 to cover the full range.

Open to feedback!

### DIFF
--- a/src/array_api_extra/__init__.py
+++ b/src/array_api_extra/__init__.py
@@ -1,10 +1,17 @@
 """Extra array functions built on top of the array API standard."""
 
-from ._delegation import expand_dims, isclose, nan_to_num, one_hot, pad, sinc
+from ._delegation import (
+    atleast_nd,
+    expand_dims,
+    isclose,
+    nan_to_num,
+    one_hot,
+    pad,
+    sinc,
+)
 from ._lib._at import at
 from ._lib._funcs import (
     apply_where,
-    atleast_nd,
     broadcast_shapes,
     cov,
     create_diagonal,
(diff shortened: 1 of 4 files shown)


lucascolley (approved):
looks great, I agree that this delegation strategy makes sense, thanks @Enderdead !

mdhaber:
@lucascolley @Enderdead `git bisect` finds that something about this commit (708bce7d24c567a1f441e893c87fa422780a5535) causes tests of `scipy.spatial.transform.RigidTransform` to fail. See e.g. first test run of https://github.com/scipy/scipy/pull/23929.

lucascolley:
@Enderdead would you be interested in opening a PR to document that we add dimensions to the front of the shape, and removing the delegation to `np.atleast_3d`?

lucascolley:
Apparently according to the docs, `np.atleast_3d` applied to a 1-D array sandwiches the existing axis between new axes 🤔 https://numpy.org/doc/2.3/reference/generated/numpy.atleast_3d.html#numpy-atleast-3d

Enderdead:
@lucascolley No problem, I’ll take care of fixing this issue. I haven’t looked into it deeply yet, but it’s concerning that the proposed unit test didn’t catch this behavior mismatch. I’ll try to push a fix by the end of the week (check unit test + remove delegation).

PR #456, opened 2025-10-02
ENH: setdiff1d: add delegation

Enderdead:
This merge request introduces the `setdiff1d` delegate function (#100).

Nothing special to mention here — it was NOT a straightforward contribution.

### DIFF
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -33,7 +33,7 @@ from array_api_extra import (
     sinc,
 )
 from array_api_extra._lib._backends import NUMPY_VERSION, Backend
-from array_api_extra._lib._testing import xp_assert_close, xp_assert_equal
+from array_api_extra._lib._testing import xfail, xp_assert_close, xp_assert_equal
 from array_api_extra._lib._utils._compat import device as get_device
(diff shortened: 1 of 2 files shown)


lucascolley:
I'm not sure about how to proceed for Dask and JAX, but I would be happy to review a PR which just adds delegation for the other libraries.

lucascolley (approved):
LGTM, thanks @Enderdead !

tests/test_funcs.py
```diff
@@ -1271,6 +1270,10 @@ def test_shapes(
     ):
         x1 = xp.zeros(shape1)
         x2 = xp.zeros(shape2)
+
+        if is_jax_namespace(xp) and assume_unique and shape1 != (1,):
+            pytest.xfail(reason="jax#32335 fixed with jax>=0.8.0")
```
crusaderky: This terminates the test early. I would like this test to XPASS (which is a failure) when we upgrade the pixi.lock, to force the maintainer to clean this up.
Instead of `pytest.xfail`, please use https://github.com/data-apis/array-api-extra/blob/ebe9a5be9826ee4c82f8aaf56c19ffa36b3a005d/src/array_api_extra/_lib/_testing.py#L279

PR #494, opened 2025-10-21
ENH: add `quantile` function with weights support

### LINKED ISSUE #340: ENH: add `quantile`
It was decided not to add `quantile` to array-api (https://github.com/data-apis/array-api/issues/795), partly because scipy added array API support to their `quantile` (https://github.com/scipy/scipy/pull/22352) and partly due to complications (many methods, weight support, nan support).

This RFC is mostly motivated by usage in scikit-learn (as seen from https://github.com/data-apis/array-api/issues/795#issuecomment-2090852426 its not used much in other packages).

Although scipy does have an array API `quantile`, due to our min scipy version, we are currently not able to use it (e.g., https://github.com/scikit-learn/scikit-learn/pull/31671).

More generally, it seems to be a commonly enough used function (which is implemented many array libraries) to warrant considering adding it to here, and having a consistent, single source.

Not sure who would be interested but maybe:
cc @betatim @ogrisel @lorentzenchr (and @rgommers who suggested it)

cakedev0:
Resolves https://github.com/data-apis/array-api-extra/issues/340

### Specs:
- **no weights:** ND-support, propagate NaNs, supported methods: `"linear", "inverted_cdf", "averaged_inverted_cdf"`
  - adding support for omitting NaNs shouldn't be too hard (if you assume NaNs are sorted at the end, which is what everyone does), but I don't think it's needed for now.
- **with weights:** Only 1D or 2D support, propagate or omit NaNs, supported methods: `"inverted_cdf", "averaged_inverted_cdf"`
- Delegates for all major supported backends when possible (numpy, cupy, torch and jax), except dask, as it was raising errors.

### Comparison with NumPy:
Those specs matches NumPy `quantile` for `nan_policy="propagate"` and `nanquantile` for `nan_policy="omit"`. Differences are:
- using `nan_policy="propagate"|"omit"` (inspired by scipy) instead of two functions like numpy (`quantile` for `nan_policy="propagate"` and `nanquantile`. Why? Less code/doc duplication, and clearer behavior IMO.
- weighted case:
  -  support for `"averaged_inverted_cdf"` method (only `"inverted_cdf"` method is supported by numpy): we need this in scikit-learn.
  - only up to 2D: we don't need more for now.
- non-weighted case:
  - less methods. Why? We don't need more. I've only found one call with a method in scikit-learn and it's a deprecated method.
  - no support for `nan_policy="omit"`: there are a few calls to `nanpercentile` in scikit-learn. Would be easy to implement, following what has been done in scipy.
  -  implementation: we sort instead of relying on partitioning like numpy does. The partitioning-based implementation would be quite more complex and would not benefits perf in most cases, as currently `xpx.partion` relies on sorting when not delegating. This could be worked

### Comparison with SciPy:

Main difference is the broadcasting/reduction behavior. We aligned on numpy for this one.

Also, SciPy doesn't support weights.

Implementation for the non-weighted case is more or less copy-pasted from SciPy, except that I've only kept support for 3 methods.

### Future use of `xpx.quantile` in scikit-learn

Those specs should be enough to replace `sklearn.utils.stats._weighted_percentile` and to replace most uses of `np.quantile/percentile`   when rewriting numpy-based functions to support Array API standard.

Note that the implementation for the weighted case differs a bit from sklearn's `_weighted_percentile`: it's mostly the same approach (sort `weights` based on `a` and compute cumulative sum), but they differ in the way edge cases are handled (null weights mostly). I believe my implementation to be easier to read and to get right, and equivalent in terms of performance (dominated by argsort for big inputs).

----

Still TODO:
- decide whether to support dask. I don't think we should it's very slow and doesn't match the purpose of dask of mentioned here. Supporting dask would slow down the CI quite a bit.
- read scikit-learn's tests for `_weighted_percentile` to see if there is something I missed in my tests


cakedev0:
This PR is now ~95% finished, so marking it ready for review. Updated PR desc.

cakedev0:
Added some comparisons with numpy/scipy in the PR description.

> Is there a link to a PR which would use this in sklearn, or are we not at that stage yet?

We are not at this stage yet, but I would be happy to open an "experiment PR" where I replace most calls to `np.percentile/np.quantile/_weighted_percentile` by calls to `xpx.quantile` to check that the tests pass. This would help a lot being confident about this PR (in terms of API decisions and implementation correctness).

lorentzenchr:
After getting weights into numpy quantile/percentile, I was so exhausted to continue pushing it elsewhere. At least now, there is one reference, numpy. I would love to see it adopted elsewhere. Quantiles are just very important in many fields.

mdhaber:
PR is up at scipy/scipy#23941. Here's a wrapper that hamstrings `stats.quantile` to get NumPy's Cartesian product behavior:

```python3
import numpy as np
from scipy import stats

def quantile_cartesian_product(x, p, method='linear', axis=0):
    p_shape = p.shape
    p = np.ravel(p)
    x = np.moveaxis(x, axis, -1)
    res = stats.quantile(x, p, method=method, axis=-1, keepdims=True)
    if len(p_shape):
        res = np.moveaxis(res, -1, 0)
        res = np.reshape(res, p_shape + res.shape[1:])
    else:
        res = np.squeeze(res, axis=-1)
    return res

# test
rng = np.random.default_rng(234892398458345982)
x = rng.random((6, 7, 8))

for axis in [-1, 0, 1]:
    for p_shape in [(), (1,), (2,), (2, 3)]:
        p = rng.random(p_shape)
        kwargs = dict(axis = 1, method='weibull')
        res = quantile_with_wonky_cartesian_product_behavior(x, p, **kwargs)
        ref = np.quantile(x, p, **kwargs)
        np.testing.assert_allclose(res, ref)
```

cakedev0:
About NEP 5: can you be more precise about how scipy's quantile respects it and not numpy's quantile?

I've read it a couple of times, and honestly I don't fully understand. One of the main example is `np.inner1d`, which is now `np.inner`. And `np.inner` does do the Cartesian product, see the [documentation](https://numpy.org/devdocs/reference/generated/numpy.inner.html): `out.shape = (*a.shape[:-1], *b.shape[:-1])`.

So to your question:

> does anything else "broadcast" by taking a Cartesian product? Imagine if this were the standard for other array operations like addition!

The answer is not as obvious as you make it sound. If the Cartesian product is valid for `np.inner` (which "NEP5-style signature" is `(i),(i)->()`) why disqualify it for `np.quantile` (which "NEP5-style signature" would be `(i),(j)->(j)` I guess)?

mdhaber:
> I've read it a couple of times, and honestly I don't fully understand. One of the main example is `np.inner1d`, which is now `np.inner`. 

"which is now `np.inner`..." - are you sure? Research (further below) suggests that `np.inner` existed before `np.inner1d`, and `inner1d` was added to meet the need for the broadcasting (rather than Cartesian product) behavior. It seems `inner1d` it was never part of the main `numpy` namespace, and it disappeared quietly[^1]. (That said, sklearn noticed: numpy/numpy#10815.)

> And `np.inner` does do the Cartesian product, see the [documentation](https://numpy.org/devdocs/reference/generated/numpy.inner.html): `out.shape = (*a.shape[:-1], *b.shape[:-1])`.

Which is diferent from what the NEP 5 says that `inner1d` did: 

>  For example, where a is of shape `(3,5,N)` and b is of shape `(5,N)`, this will return an output of shape `(3,5)`

```python3
import numpy as np
N = 10
a = np.ones((3, 5, N))
b = np.ones((5, N))
np.inner(a, b).shape  # (3, 5, 5)
```

I'd suggest that `matmul` and `vecdot` *correct* the behavior of `dot` and `inner`. For instance, from `matmul`:

> The matmul function implements the semantics of the `@` operator introduced in Python 3.5 following [PEP 465](https://peps.python.org/pep-0465/).

Which states:
> For inputs with more than 2 dimensions, we treat the last two dimensions as being the dimensions of the matrices to multiply, and ‘broadcast’ across the other dimensions.

`np.vecdot` was added because broadcasting, rather than Cartesian product, is the behavior offered by the Array API standard [`vecdot`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.vecdot.html) ("shape determined according to Broadcasting along the non-contracted axes").

I say "correct" because presumably the discussions that led to the inclusion of the `@` operator in Python (PEP 465) and the array API were more careful than whatever led to `np.inner`, which has had the Cartesian product behavior since at least [NumPy 1.3](https://numpy.org/doc/stable/release/1.3.0-notes.html#generalized-ufuncs). Or so I infer from [this mailing list discussion](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/DLUL2KZAMFRGOAHS4DLA2GGDMYL4MFBB/#DLUL2KZAMFRGOAHS4DLA2GGDMYL4MFBB), which seems to have been the precursor to NEP 5. Since then, there have been discussions started by users looking for the broadcasting behavior and not getting what they wanted from `inner`, e.g. [here](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/XMIICHAQ5PQXNW5GEF5HVA6KWK4UURCC/#QT6E4WZTPQEPWFFJDEEJ2RGSDZI62TMA) and [here](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/VRHLD7AEP5FMXAZ4GEON25QXWZ7MG5TL/#PGDHJLS35NBO73NGJJIFITXSPCIIKUWQ).

cakedev0:
... Well that is very convincing. Immense thank you for the detailed explanation, I definitely learned a lot there (and sorry for the time you had to invest, I hope I'll be able to make good use of the understanding I gained here in future contribution).

> which is now np.inner

Indeed no. Sorry, I assumed that based on the very similar names, my bad!

cakedev0:
I still think that this behavior isn't right:

> both quantile([0, 1], [0.5]) and quantile([0, 1], 0.5) have dimension 0 but quantile([0, 1], [0.5, 0.9]) has dimension 1.

But this is more of a detail, and I do understand why it's like this. I'll try to see if I can propose something I would find better, but I'm not sure I'll be able to.

Maybe no `keepdims`? The signature is `(i),(j)->(j)`, it doesn't reduce away a dimension, so maybe it's fair to not have a `keepdims` parameter? I could not find a gufunc with such signature in numpy... For `matmul`, passing `keepdims` is forbidden, but the error message says it's because it's 2d.

You said you don't like `keepdims`, so maybe you'll like this idea :wink:

cakedev0:
Sorry for the long delay. I must say I loss most of my motivation on this topic... @lucascolley feel free to close the PR and @mdhaber feel free to not respond.

Basically, I mostly agree with everything in your previous comment. Especially with: 

> So you want to get the best of both worlds - efficiency benefits of (i),(j)->(j) while respecting the fact that the quantile operation is naturally (i),()->()

I think that I would have chosen a behavior like this instead of what scipy did:
- if `q` is scalar do `(i),()->()`
- otherwise do `(i),(j)->(j)` (even for `j = 1`)
- no `keepdims`.

To me this looks familiar and expected as it's close to `numpy.matmul` / `@` operator behavior: `(n) @ (n) -> ()`, `(m, n) @ (n) -> (m)` (even for `m = 1`), `(m, n) @ (n, k) -> (m, k)` (even for `m = k = 1`).

> I'm not sure if changing that is worth a deprecation cycle or a separate function

I don't know what's the cost of a deprecation cycle all I can say is that I really don't like that:

> both quantile([0, 1], [0.5]) and quantile([0, 1], 0.5) have dimension 0 but quantile([0, 1], [0.5, 0.9]) has dimension 1.

Even if you can control it with `keepdims`. For me it breaks a strong expectation I have about functions on arrays. But that's personal, and even if you agree you might still judge it's not worth a deprecation cycle.

cakedev0:
Closing this PR:
- I won't have the bandwidth to go back on it anytime soon
- SciPy has quantile now (and it's very powerful)
- We'll maintain our own version in scikit-learn, until we bump scipy min version to 1.17, then we'll rely on scipy's quantile.

Thank you for the feedback, and sorry if I was a bit stubborn.

lucascolley:
sounds good, thanks for driving the discussion @cakedev0 !

PR #496, opened 2025-10-23
MAINT: Pixi maintenance

crusaderky:
- Bump pixi to latest version
- Fix issue that causes pixi 0.58 to fail when it's called from lefthook in parallel mode. I observed the same problem sporadically with past versions too.
- Run `pixi update`. No noteworthy changes.
- Misc CI tweaks

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -1,4 +1,7 @@
-lefthook: pixi run --no-progress --environment=lint lefthook
+lefthook: pixi run --no-progress --environment=lint --frozen lefthook
+
+templates:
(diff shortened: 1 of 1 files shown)


lucascolley:
Interesting. @hofer-julian do you know if there is an open issue for this?

Pixi's CI itself is still using this lefthook setup, so maybe there is another difference that means this isn't hit over there? What are the files created/deleted?

The disadvantage is that the duplication of lint task definitions increases maintenance burden a little.

Hofer-Julian:
Maybe we should add --frozen for now? I guess this would at least avoid this problem for now

Issue #502, opened 2025-10-28
ENH: vectorize `cov`


mdhaber:
`cov` currently expects an array with at most two dimensions.  Given a single $m$-dimensional sample composed of $n$ observations as an array with shape `(m, n)`, it produces one $m \times m$ covariance matrix as an array of shape `(m, m)`.

It would be useful if `cov` were vectorized operate on arrays of arbitrary dimensionality as a batch. Given a batch of samples as an array with shape `batch_shape + (m, n)`, it would produce a batch of covariance matrices as an array of shape `batch_shape + (m, m)`. 

This would be consistent with other standard functions like `matmul`. `matrix_transpose`, and many `linalg` extension functions.

(I'd have an immediate use for this vectorizing `scipy.stats.epps_singleton_2samp`.)

Issue #324, opened 2025-06-05
API: decide on a backwards-compat policy


lucascolley:
Putting a 1.0 milestone on this.

crusaderky:
I think there should be a single policy between array-api-compat and array-api-extra.
Probably this thread should be in `/array-api/`?

lucascolley:
will link to that thread when it exists 👍

MarcoGorelli:
Just for reference / in case it's useful, here's what we adopted in Narwhals: [Perfect backwards compatibility policy](https://narwhals-dev.github.io/narwhals/backcompat/)

Or, if you prefer video/audio: https://youtu.be/r2PxJlO7_QA?feature=shared&t=849

lucascolley:
we discussed this in the community meeting and the Narwhals policy looks great as an end-goal. For now, the current no guarantees and recommendation to vendor serves the project well. Perhaps some middle-ground would be appropriate to adopt in the not-so-distant future.

PR #507, opened 2025-10-31
ENH: vectorize `cov`

mdhaber:
Closes gh-502

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -157,7 +157,6 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
 
            [[ 46.84      , -17.144     ],
             [-17.144     ,   8.57653333]]], dtype=array_api_strict.float64)
-
     """
 
     if xp is None:
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -480,10 +480,10 @@ class TestCov:
         rng = np.random.default_rng(8847643423)
         batch_shape = (3, 4)
         n_var, n_obs = 3, 20
-        m = rng.random(batch_shape + (n_var, n_obs))
+        m = rng.random((*batch_shape, n_var, n_obs))
         res = cov(xp.asarray(m))
-        ref = [np.cov(m_) for m_ in np.reshape(m, (-1, n_var, n_obs))]
-        ref = np.reshape(np.stack(ref), batch_shape + (n_var, n_var))
+        ref_list = [np.cov(m_) for m_ in np.reshape(m, (-1, n_var, n_obs))]
+        ref = np.reshape(np.stack(ref_list), (*batch_shape, n_var, n_var))
         xp_assert_close(res, xp.asarray(ref))
 
 


src/array_api_extra/_delegation.py
```diff
@@ -142,6 +146,18 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    two-dimensional input.
+
+    >>> stack = xp.stack((X, 2*X))
+    >>> xpx.cov(stack)
+    Array([[[ 11.71      ,  -4.286     ],
+            [ -4.286     ,   2.14413333]],
+
+           [[ 46.84      , -17.144     ],
+            [-17.144     ,   8.57653333]]], dtype=array_api_strict.float64)
+
```
mdhaber:
```suggestion
```

tests/test_funcs.py
```diff
@@ -476,6 +476,16 @@ def test_xp(self, xp: ModuleType):
...
         )
 
+    def test_batch(self, xp: ModuleType):
+        rng = np.random.default_rng(8847643423)
+        batch_shape = (3, 4)
+        n_var, n_obs = 3, 20
+        m = rng.random(batch_shape + (n_var, n_obs))
+        res = cov(xp.asarray(m))
+        ref = [np.cov(m_) for m_ in np.reshape(m, (-1, n_var, n_obs))]
+        ref = np.reshape(np.stack(ref), batch_shape + (n_var, n_var))
```
mdhaber:
```suggestion
        ref = np.reshape(np.stack(ref_list), batch_shape + (n_var, n_var))
```

mdhaber (commented):
Should address the lint failures.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -258,20 +258,20 @@ def cov(m: Array, /, *, xp: ModuleType) -> Array:  # numpydoc ignore=PR01,RT01
...
     m_shape = eager_shape(m)
-    fact = m_shape[1] - 1
+    fact = m_shape[-1] - 1
 
     if fact <= 0:
         warnings.warn("Degrees of freedom <= 0 for slice", RuntimeWarning, stacklevel=2)
         fact = 0
 
-    m -= avg[:, None]
-    m_transpose = m.T
```
lucascolley: can't we use https://data-apis.org/array-api/draft/API_specification/generated/array_api.array.mT.html#array_api.array.mT ?

mdhaber: That's not implemented in all backends yet (at least not CuPy and dask.array on colab) and it can't be added by `array_api_compat`.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -258,20 +258,20 @@ def cov(m: Array, /, *, xp: ModuleType) -> Array:  # numpydoc ignore=PR01,RT01
...
+    m -= avg
+    m_transpose = xp.matrix_transpose(m)
     if xp.isdtype(m_transpose.dtype, "complex floating"):
         m_transpose = xp.conj(m_transpose)
-    c = m @ m_transpose
+    c = xp.matmul(m, m_transpose)
```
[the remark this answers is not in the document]
mdhaber: TBH I didn't know they were supposed to be equivalent until after looking it up. I wasn't sure what `@` did for n-D arrays, and `matmul` is a lot easier to look up 

If @ really does work like it should for all backends, it can be left alone. Do you know? If not, safer to use the function because it can be updated by `array_api_compat`.

tests/test_funcs.py
```diff
@@ -476,6 +476,16 @@ def test_xp(self, xp: ModuleType):
             xp.asarray([[1.0, -1.0], [-1.0, 1.0]], dtype=xp.float64),
         )
 
+    def test_batch(self, xp: ModuleType):
+        rng = np.random.default_rng(8847643423)
+        batch_shape = (3, 4)
+        n_var, n_obs = 3, 20
+        m = rng.random(batch_shape + (n_var, n_obs))
+        res = cov(xp.asarray(m))
+        ref = [np.cov(m_) for m_ in np.reshape(m, (-1, n_var, n_obs))]
```
mdhaber:
```suggestion
        ref_list = [np.cov(m_) for m_ in np.reshape(m, (-1, n_var, n_obs))]
```

mdhaber (commented):
BTW @lucascolley, while we were at it, I wanted to ask about 1) enforcing use of double precision and 2) squeezing all singleton dimensions at the end.

I've heard it argued that double precision arithmetic is essential for accurate covariance calculations.
Double precision arithmetic is essentially *always* important for "accurate" calculations. Personally, I've never used `float32` intentionally, but I understand it's important, and I don't see why fundamental libraries should make the choice for the user.

As for the "indescriminate squeeze", I can see why you *might* want to eliminate a dimension if the covariance between two univariate samples were viewed as a reducing operation. But that calculation is not really possible with one-argument `cov`, and there is no reason to use `cov` to take the variance of a single univariate sample. So really, I don't see any good use case for 1D input anyway, and for any other dimensionality, it's not good to eliminate singleton batch dimensions.

lucascolley:
> Does `array_api_extra` need deprecation and all that

currently TBC: https://github.com/data-apis/array-api-extra/issues/324

I'd probably be happy to ship the proposed changes in a 0.10.0, supposing no further movement on that issue.

PR #514, opened 2025-11-06
BUG: Fix delegation behaviour with `atleast_3d`

Enderdead:
This PR fixes the bad behavior reported in scipy/scipy#23929 when `ndim`=3.

It removes delegation when `ndim` > 2 and adds additional tests.
These tests reproduce the issue while avoiding input arrays of size 1.

I'm not entirely happy with these new tests, but they provide a reasonable starting point for further iteration.

Feedback is welcome.

### DIFF
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -311,7 +311,7 @@ class TestAtLeastND:
         xp: ModuleType,
     ):
         n = math.prod(input_shape)
-        x = xp.reshape(xp.arange(n), input_shape)
+        x = xp.asarray(np.arange(n).reshape(input_shape))
         y = atleast_nd(x, ndim=ndim)
 
         assert y.shape == expected_shape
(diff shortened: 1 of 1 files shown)


Enderdead:
@lucascolley  I’m proposing this new version with a hash method to perform a semi–sanity check.
Sorry if it’s still a bit heavy — it’s difficult for me to gauge how far we can go with simplifying this test.

tests/test_funcs.py
```diff
@@ -291,7 +293,31 @@ def test_0D(self, xp: ModuleType):
...
+    )
+    def test_1D_shapes(
+        self,
+        input_shape: tuple[int],
+        ndim: int,
+        expected_shape: tuple[int],
+        xp: ModuleType,
+    ):
+        n = math.prod(input_shape)
+        x = xp.reshape(xp.asarray(list(range(n))), input_shape)
```
lucascolley: can `xp.arange` be used in place of the asarray list range combo here?

lucascolley:
@allcontributors please add @Enderdead for bug

PR #531, opened 2025-11-23
ENH: Add vectorized `searchsorted`

mdhaber:
Addresses https://github.com/scipy/scipy/pull/23930#issuecomment-3488593722

Thoughts:
- Currently this delegates to the backend's native `searchsorted` for all 1D tests. It would be good to have some way of running the 1D test cases with the generic `searchsorted` code, though. The important part is the length of the arrays along the last axis, so the simplest way would be to just define 2D cases with the same lengths along the second axis. 
- `torch` delegates multidimensional cases to `torch.searchsorted`, too, but it behaves differently than other backends in some cases involving NaNs. It doesn't surprise me because the standard doesn't define sorting w/ NaNs and torch doesn't always do the typical thing. Not sure how deeply we want to go into that.
- Currently, arrays should have identical shapes except along the last axis; otherwise the backends will typically raise their own errors at some point. In SciPy, we just broadcast the batch dimensions, but I didn't want to port all that code over here. It's possible for someone to adapt `broadcast_shapes` to help with that, but I would prefer not to work on that. I could add a check for this condition and raise a consistent error, if that would be sufficient.
- LMK what other input validation array-api-extra would expect - e.g. whether `side` is one of the allowed values?

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -647,9 +647,9 @@ def searchsorted(
     Find the indices into a sorted array ``x1`` such that if the elements in ``x2``
     were inserted before the indices, the resulting array would remain sorted.
 
-    The behavior of this function is similar to that of the homonymous function in the
-    array API standard, but it relaxes the requirement that `x1` must be
-    one-dimensional. The function is vectorized, treating slices along the last axis
+    The behavior of this function is similar to that of `array_api.searchsorted`,
+    but it relaxes the requirement that `x1` must be one-dimensional.
+    This function is vectorized, treating slices along the last axis
     as elements and preceding axes as batch (or "loop") dimensions.
 
     Parameters
@@ -701,11 +701,11 @@ def searchsorted(
         raise ValueError(message)
 
     xp_default_int = _funcs.default_dtype(xp, kind="integral")
-    y_0d = xp.asarray(x2).ndim == 0
-    x_1d = x1.ndim <= 1
+    x2_0d = x2.ndim == 0
+    x1_1d = x1.ndim <= 1
 
-    if x_1d or is_torch_namespace(xp):
-        x2 = xp.reshape(x2, ()) if (y_0d and x_1d) else x2
+    if x1_1d or is_torch_namespace(xp):
+        x2 = xp.reshape(x2, ()) if (x2_0d and x1_1d) else x2
         out = xp.searchsorted(x1, x2, side=side)
         return xp.astype(out, xp_default_int, copy=False)
 
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -1789,7 +1789,7 @@ def xp_searchsorted(
     side: Literal["left", "right"],
     xp: ModuleType,
 ) -> Array:
-    return xp.searchsorted(xp.asarray(a), xp.asarray(v), side=side)
+    return xp.searchsorted(a, v, side=side)
 
 
 @pytest.mark.skip_xp_backend(Backend.DASK, reason="no take_along_axis")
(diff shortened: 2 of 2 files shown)


tests/test_funcs.py
```diff
@@ -1637,3 +1643,157 @@ def test_kind(self, xp: ModuleType, library: Backend):
...
+        y = rng.permuted(np.concatenate((xl, x, xr, x_), axis=-1), axis=-1)
+        if nans_x:
+            mask = rng.random(shape) < 0.1
+            x[mask] = np.nan
+        if infs_x:
+            mask = rng.random(shape) < 0.1
+            x[mask] = -np.inf
+            mask = rng.random(shape) > 0.9
+            x[mask] = np.inf
+        x = np.sort(x, stable=True, axis=-1)
```
mdhaber: This is unnecessary and causing CI failures.
```suggestion
        x = np.sort(x, axis=-1)
```

lucascolley:
> Re: pyright failures - I see 144 errors in main locally. Am I supposed to look past those to fix the errors in the new code?

How are you seeing these failures? Via `pixi run pyright`?

lucascolley:
seems to be the problem yes, I fear it might be something windows-specific.

lucascolley:
let me know if you would rather get this into the next release or wait for the one after

mdhaber:
@lucascolley let's get this into this release. Could you please fix the linter issue? I'm not getting that locally, and type-checking unit tests is beyond my capabilities as a programmer.

src/array_api_extra/_delegation.py
```diff
@@ -632,6 +633,85 @@ def pad(
...
+    """
+    Find indices where elements should be inserted to maintain order.
+
+    Find the indices into a sorted array ``x1`` such that if the elements in ``x2``
+    were inserted before the indices, the resulting array would remain sorted.
+
+    The behavior of this function is similar to that of the homonymous function in the
+    array API standard, but it relaxes the requirement that `x1` must be
+    one-dimensional. The function is vectorized, treating slices along the last axis
+    as elements and preceding axes as batch (or "loop") dimensions.
```
lucascolley: Thanks. Non-blocking since our coverage of this is patchy elsewhere, but bonus points for also inter-sphinxing with the standard, like https://github.com/data-apis/array-api-extra/blob/763ea1a76675b40443e6f45751be3f0ddf389662/src/array_api_extra/_delegation.py#L453-L455

tests/test_funcs.py
```diff
@@ -53,6 +62,8 @@
 lazy_xp_function(pad)
 # FIXME calls in1d which calls xp.unique_values without size
 lazy_xp_function(setdiff1d, jax_jit=False)
+lazy_xp_function(xpx_searchsorted)
+lazy_xp_function(_funcs_searchsorted)
```
lucascolley: as a sidenote I wonder whether we are missing some of these (e.g. `default_dtype`)

src/array_api_extra/_delegation.py
```diff
@@ -632,6 +633,85 @@ def pad(
...
+    if xp is None:
+        xp = array_namespace(x1, x2)
+
+    if side not in {"left", "right"}:
+        message = "`side` must be either 'left' or 'right'."
+        raise ValueError(message)
+
+    xp_default_int = _funcs.default_dtype(xp, kind="integral")
+    y_0d = xp.asarray(x2).ndim == 0
+    x_1d = x1.ndim <= 1
```
mdhaber:
```suggestion
    x1_1d = x1.ndim <= 1
```

mdhaber:
Thanks. Hopefully got it all this time.

lucascolley (approved):
🚀

Issue #532, opened 2025-11-23
xpx.create_diagonal was broken for broadcasting


NeilGirdhar:
```
[utM] In [1]: import array_api_extra as xpx
[utM] In [2]: import numpy as np
[utM] In [4]: x = np.zeros((10, 4))
[utM] In [5]: xpx.create_diagonal(x)
```
Crashes.

@Enderdead Not sure if it was the delegation?

NeilGirdhar:
Yeah, looks like the `<3` condition that was added in the "delegation" change (seems to do more than just delegation) is trying to implement `create_diagonal` using functions that can't do it for 2-dimensional arrays.

Easy fix is probably to switch to <2, and probably write a test to make sure this doesn't get tripped again?

lucascolley:
Is the problem what `np.diag` returns in the 2-D case?

PR #544, opened 2025-12-07
BUG: create_diagonal: remove delegation for `ndim >= 2`

Enderdead:
I applied the quick fix (#532) to the create_diagonal delegation function by delegating only when the dimension is less than 2.

I also added another test case to capture the bug, and removed an unused type that I forgot to clean up in a previous MR.


lucascolley (approved):
thanks @Enderdead !

PR #582, opened 2026-01-03
ENH: testing.lazy_xp_function: method, classmethod, staticmethod inheritance support

### LINKED ISSUE #488: ENH: Add support for something like `lazy_xp_function` for classes
`lazy_xp_function` is quite effective for testing functions with backends that support lazy behavior. SciPy  uses it within [`make_xp_pytest_marks` ](https://github.com/scipy/scipy/blob/71fb82e33c47a4358804578a6a16ed9129ef4049/scipy/_lib/_array_api.py#L896) which generates the pytest marks for which backends to run tests on. 

The PR https://github.com/scipy/scipy/pull/23135 attempts to adapt SciPy's strategy for testing functions against alternative Python Array API standard compatible backends to some classes, including [PPoly](https://github.com/scipy/scipy/blob/4622f22fc613341696b7cdfaba9fb0a46515bbfa/scipy/interpolate/_interpolate.py#L841-L842), and things almost work out of the box, with the exception of testing on lazy backends. This is because on dask `lazy_xp_function` wraps its input with [_dask_wrap](https://github.com/data-apis/array-api-extra/blob/20593693e22bcfc0fddee27fcacc91abc8c8dbf7/src/array_api_extra/testing.py#L382), which replaces the class with a function (something similar happens with JAX and `jax_autojit`) This resulted in the mysterious error here, https://github.com/scipy/scipy/pull/23135#issuecomment-3265704061.

For the time being, I think it's fine to just skip testing classes and their methods against lazy backends in SciPy, but longer term, it would be good to think about what should be done here. `lazy_xp_class` if it existed, could call `_dask_wrap` individually on methods. There's a question of whether you'd want separate `CountingDaskScheduler`s per method, or one shared one per class or instance. I don't have strong opinions about how exactly things are done and haven't thought about it closely. I'm mainly making this issue to give somewhere to point to explain why classes in SciPy aren't being tested for lazy behavior, even if everything actually works, and to raise awareness.

steppi:
Closes #488 

In #488 I asked about applying something like `lazy_xp_function` to classes. This was a half-baked idea motivated by wanting `xp_capabilities` in SciPy, which everyone here should be familiar with, to always be applied at the class level instead of the individual method level. In https://github.com/scipy/scipy/pull/24267 I created a workflow that keeps `xp_capabilities` applied to classes, but still allows capabilities to be defined separately for individual methods if needed, and allows `lazy_xp_function` to be applied to individual methods separately. This means no nonsense about trying to automatically determine which methods of a class should have `lazy_xp_function` applied to them.

While working on this, I settled on specifying methods with tuples of the form `Tuple[type, str]` specifying an (uninstantiated) class and a method name. This is to allow distinguishing things like `A.f` from `B.f` when `B` is a subclass of `A` that inherits `f` from `A`, since capabilities may differ at different levels of the inheritance hierarchy. Through changes in SciPy, I was able to allow precise declarations of capabilities for class methods in the presence of inheritance, but a separate change is needed here to allow things like applying `lazy_xp_function` to `B.f` but not `A.f` when `f` is inherited from `A`.

The change here modifies `lazy_xp_function` to also take tuples of the form `Tuple[type, str]`. When this is done, for say `(B, "f")`, then `B.f` is replaced with a shallow clone of itself before adding the tags. This allows`B.f` gets the tags without `A.f` to get them. If replacing an inherited method with a shallow clone is too obtrusive, I have a workaround that keeps more obtrusive modifications only within `patch_lazy_xp_functions`, but it makes things considerably more complicated, so I hope that won't be necessary.

https://github.com/scipy/scipy/blob/dfa1b87e4af7cf7ee5a8b8faf5c4360b63c86b36/scipy/_lib/tests/test_xp_capabilities.py has an (xfailed) test giving an example of what can go wrong with inheritance currently which can be made to pass with the change made in this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "mesonpy"
 
 [project]
 name = "array-api-extra"
-version = "0.10.0.dev0"
+version = "0.10.1.dev0"
 authors = [
   { name = "Lucas Colley", email = "lucas.colley8@gmail.com" },
   # { name = "Open Source Contributors" },  # https://github.com/pypi/warehouse/issues/14813
@@ -23,6 +23,7 @@ classifiers = [
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
   "Programming Language :: Python :: Free Threading :: 3 - Stable",
   "Typing :: Typed",
 ]
@@ -56,20 +57,20 @@ array-api-compat = "*"
 ### workspace environments ###
 
 [tool.pixi.environments]
-default = { features = ["py313"], solve-group = "py313" }
-lint = { features = ["py313", "lint"], solve-group = "py313" }
-docs = { features = ["py313", "docs"], solve-group = "py313" }
-tests = { features = ["py313", "tests"], solve-group = "py313" }
-tests-py313 = { features = ["py313", "tests"], solve-group = "py313" } # alias of tests
+default = { features = ["py314"], solve-group = "py314" }
+lint = { features = ["py314", "lint"], solve-group = "py314" }
+docs = { features = ["py314", "docs"], solve-group = "py314" }
+tests = { features = ["py314", "tests"], solve-group = "py314" }
+tests-py314 = { features = ["py314", "tests"], solve-group = "py314" } # alias of tests
 
 # Some backends may pin numpy; use separate solve-group
-dev = { features = ["py313", "lint", "tests", "docs", "dev", "backends"], solve-group = "backends" }
-tests-backends = { features = ["py313", "tests", "backends"], solve-group = "backends" }
+dev = { features = ["py314", "lint", "tests", "docs", "dev", "backends"], solve-group = "backends" }
+tests-backends = { features = ["py314", "tests", "backends"], solve-group = "backends" }
 tests-backends-py311 = { features = ["py311", "tests", "backends"] }
 
 # CUDA not available on free github actions and on some developers' PCs
-dev-cuda = { features = ["py313", "lint", "tests", "docs", "dev", "backends", "cuda-backends"], solve-group = "cuda" }
-tests-cuda = { features = ["py313", "tests", "backends", "cuda-backends"], solve-group = "cuda" }
+dev-cuda = { features = ["py314", "lint", "tests", "docs", "dev", "backends", "cuda-backends"], solve-group = "cuda" }
+tests-cuda = { features = ["py314", "tests", "backends", "cuda-backends"], solve-group = "cuda" }
 tests-cuda-py311 = { features = ["py311", "tests", "backends", "cuda-backends"] }
 
 # Ungrouped environments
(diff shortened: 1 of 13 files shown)


lucascolley:
cc @crusaderky

steppi:
Thanks @lucascolley. I've pushed a commit which fixes the failures locally.

steppi:
> Are they hit by `pixi run -e tests-numpy1 tests-ci` or not? If not, is there something else than just the NumPy version?

Interesting, I tried it locally and it seems like they are still not hit. But they should be hit by this test if an older NumPy is installed::

```python
try:
    # Test an arbitrary Cython ufunc (@cython.vectorize).
    # When SCIPY_ARRAY_API is not set, this is the same as
    # scipy.special.erf.
    from scipy.special._ufuncs import erf  # type: ignore[import-untyped]

    lazy_xp_function(erf)
except ImportError:
    erf = None


@pytest.mark.skip_xp_backend(Backend.TORCH_GPU, reason="device->host copy")
@pytest.mark.filterwarnings("ignore:__array_wrap__:DeprecationWarning")  # PyTorch
def test_lazy_xp_function_cython_ufuncs(xp: ModuleType, library: Backend):
    pytest.importorskip("scipy")
    assert erf is not None
    x = xp.asarray([6.0, 7.0])
    if library.like(Backend.ARRAY_API_STRICT, Backend.JAX):
        # array-api-strict arrays are auto-converted to NumPy
        # which results in an assertion error for mismatched namespaces
        # eager JAX arrays are auto-converted to NumPy in eager JAX
        # and fail in jax.jit (which lazy_xp_function tests here)
        with pytest.raises((TypeError, AssertionError)):
            xp_assert_equal(cast(Array, erf(x)), xp.asarray([1.0, 1.0]))
    else:
        # CuPy, Dask and sparse define __array_ufunc__ and dispatch accordingly
        # note that when sparse reduces to scalar it returns a np.generic, which
        # would make xp_assert_equal fail.
        xp_assert_equal(cast(Array, erf(x)), xp.asarray([1.0, 1.0]))
```

steppi:
> Wanna try `pixi add -f tests scipy`? If there are any difficulties then probably fine to just ignore it.

I can confirm that adding scipy causes those lines to get hit locally.

src/array_api_extra/testing.py
```diff
@@ -209,10 +223,21 @@ def test_myfunc(xp):
         "jax_jit": jax_jit,
     }
 
+    if isinstance(func, tuple):
+        # Replace the method with a clone before adding tags
+        # to avoid adding unwanted tags to a parent method when
+        # the method was inherited from a parent class.
+        cls, method_name = func
```
crusaderky:
```suggestion
        # Note: can't just accept an unbound method `cls.method_name` because in
        # case of inheritance it would be impossible to attribute it to the child class.
        cls, method_name = func
```

tests/test_testing.py
```diff
@@ -321,6 +322,49 @@ def test_lazy_xp_function_cython_ufuncs(xp: ModuleType, library: Backend):
...
+        assert not hasattr(A.g, "_lazy_xp_function")
+
+    def test_lazy_xp_function_classes(self, xp):
+        x = xp.asarray([1.1, 2.2, 3.3])
+        y = xp.asarray([1.0, 2.0, 3.0])
+        z = xp.asarray([3.0, 4.0, 5.0])
+        foo = B(x)
+        observed = foo.g(y, z)
+        expected = xp.asarray(44.0)[()]
+        xp_assert_close(observed, expected)
```
crusaderky: This runs the function, but it doesn't test that it's been wrapped.
You need to write another function that will fail when wrapped, e.g.
```python
def w(self):
    return bool(self._xp.any(self.x))`
```
See tests earlier in this same test module for examples.

crusaderky (changes requested):

steppi:
Thanks @crusaderky; the test is actually doing what it's supposed to now.

steppi:
Phew, the classmethod and staticmethod situation was more complicated than I realized, and these actually weren't getting wrapped before this PR either if someone did something like `lazy_xp_function(scipy.spatial.Rotation.from_quat)` like what happens here:

https://github.com/scipy/scipy/blob/10bd7ba5e1b7a537856e08c2168796342fb053e7/scipy/spatial/transform/tests/test_rotation.py#L35.

They get wrapped now and I've added a test to confirm it. Still thinking about what to do with cases like:

```python
class eager:
    # Bypass xpx.testing.lazy_xp_function when calling
    # these functions from this namespace
    is_valid_im = is_valid_im
    is_valid_linkage = is_valid_linkage
```

This behavior is documented, so it needs to still work

https://github.com/data-apis/array-api-extra/blob/b6518f30a4e7205c26a90a47edfd2577f5773ecf/src/array_api_extra/testing.py#L183-L196

steppi:
I figured something out. We can use the `__qualname__` to distinguish methods that belong to a class  from references to functions that are defined elsewhere. I already made it so `lazy_xp_function` replaces methods with clones, and now it also updates the `__qualname__`, making it so the `__qualname__` check can still works when there is inheritance.

pyproject.toml
```diff
@@ -133,6 +133,7 @@ pytest-cov = ">=7.0.0"
 hypothesis = ">=6.148.8"
 array-api-strict = ">=2.4.1,<2.5"
 numpy = ">=1.22.0"
+scipy = ">=1.15.2,<2"
```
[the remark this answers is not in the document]
lucascolley: It was deliberate for test coverage. I can handle conflicts

PR #690, opened 2026-04-17
ENH: cov: expose correction and weights parameters

### LINKED ISSUE #688: EHN: make covariance more flexible
At the moment, the covariance function from the array api extra does not allow the use of any parameters, such as bias/ddof/fweights/aweights/rowvar.

These numpy and torch parameters make a difference for many applications with covariance. I was wondering if it would be okay to expose and normalize the parameters. It is a really simple PR, a few lines.

bruAristimunha:
Resolves #688.

## Summary

- Adds `axis`, `correction`, `frequency_weights`, and `weights` parameters to `xpx.cov`, unlocking the degrees-of-freedom and weighted variants that `numpy.cov` and `torch.cov` already support.
- Naming follows array-api conventions (`axis`, `correction`) used elsewhere in this library rather than numpy's (`rowvar`, `bias`, `ddof`). The docstring includes a one-to-one mapping for users migrating from `numpy.cov`.

## Design

The delegation moves observations to the last axis via `xp.moveaxis`, which collapses `rowvar` out of backend dispatch entirely — only `ddof` (numpy/cupy/dask/jax) vs `correction` (torch) differs between branches.

Fallbacks to the generic implementation (`_funcs.cov`):
- `m.ndim > 2` (batched input, not supported by any native).
- Non-integer `correction` (rejected by `numpy.cov`'s `ddof`).
- Dask with weights — `dask.array.cov` forces `.compute()` on a lazy 0-D scalar via its internal `if fact <= 0` check. The generic path stays fully lazy because its weighted branch doesn't compare `fact` to zero (noted in docstring).

Weighted formula in `_funcs.cov` matches numpy's (algebraically): `c = (m_c · w) @ m_c.T / (v1 - correction · v2 / v1)`.

## Tests

New `TestCov` cases validate against `np.cov` as reference:
- `test_correction` (integer ddof)
- `test_correction_float` (generic-path-only, hand-computed reference)
- `test_axis` / `test_axis_with_weights` / `test_axis_out_of_bounds`
- `test_frequency_weights` / `test_weights` / `test_both_weights`
- `test_batch_with_weights`

## Test plan
- [x] `pytest tests/test_funcs.py::TestCov` — 126 passed across numpy, torch, jax, dask, array-api-strict
- [x] `pytest tests/test_funcs.py` full — 4263 passed, 0 failed
- [x] `lefthook run pre-commit` — ruff, numpydoc, mypy, pyright, typos all green
- [x] Dask laziness verified — `lazy_xp_function(cov)` asserts 0 `.compute()` calls, holds for weighted path via the fallback

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -81,7 +81,16 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
     return _funcs.atleast_nd(x, ndim=ndim, xp=xp)
 
 
-def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
+def cov(
+    m: Array,
+    /,
+    *,
+    axis: int = -1,
+    correction: int | float = 1,
+    frequency_weights: Array | None = None,
+    weights: Array | None = None,
+    xp: ModuleType | None = None,
+) -> Array:
     """
     Estimate a covariance matrix (or a stack of covariance matrices).
 
@@ -92,16 +101,37 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
     of :math:`x_i`.
 
-    With the exception of supporting batch input, this provides a subset of
-    the functionality of ``numpy.cov``.
+    Extends ``numpy.cov`` with support for batch input and array-api
+    backends. Naming follows the array-api conventions used elsewhere in
+    this library (``axis``, ``correction``) rather than the numpy spellings
+    (``rowvar``, ``bias``, ``ddof``); see Notes for the mapping.
 
     Parameters
     ----------
     m : array
         An array of shape ``(..., N, M)`` whose innermost two dimensions
-        contain *M* observations of *N* variables. That is,
-        each row of `m` represents a variable, and each column a single
-        observation of all those variables.
+        contain *M* observations of *N* variables by default. The axis of
+        observations is controlled by `axis`.
+    axis : int, optional
+        Axis of `m` containing the observations. Default: ``-1`` (the last
+        axis), matching the array-api convention. Use ``axis=-2`` (or ``0``
+        for 2-D input) to treat each column as a variable, which
+        corresponds to ``rowvar=False`` in ``numpy.cov``.
+    correction : int or float, optional
+        Degrees of freedom correction: normalization divides by
+        ``N - correction`` (for unweighted input). Default: ``1``, which
+        gives the unbiased estimate (matches ``numpy.cov`` default of
+        ``bias=False``). Set to ``0`` for the biased estimate (``N``
+        normalization). Corresponds to ``ddof`` in ``numpy.cov`` and to
+        ``correction`` in ``numpy.var``/``std`` and ``torch.cov``.
+    frequency_weights : array, optional
+        1-D array of integer frequency weights: the number of times each
+        observation is repeated. Corresponds to ``fweights`` in
+        ``numpy.cov``/``torch.cov``.
+    weights : array, optional
+        1-D array of observation-vector weights (analytic weights). Larger
+        values mark more important observations. Corresponds to
+        ``aweights`` in ``numpy.cov``/``torch.cov``.
     xp : array_namespace, optional
         The standard-compatible namespace for `m`. Default: infer.
 
@@ -110,6 +140,23 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     array
         An array having shape (..., N, N) whose innermost two dimensions represent
         the covariance matrix of the variables.
+
+    Notes
+    -----
+    Mapping from ``numpy.cov`` to this function::
+
+        numpy.cov(m, rowvar=True)           -> cov(m, axis=-1)   # default
+        numpy.cov(m, rowvar=False)          -> cov(m, axis=-2)
+        numpy.cov(m, bias=True)             -> cov(m, correction=0)
+        numpy.cov(m, ddof=k)                -> cov(m, correction=k)
+        numpy.cov(m, fweights=f)            -> cov(m, frequency_weights=f)
+        numpy.cov(m, aweights=a)            -> cov(m, weights=a)
+
+    Unlike ``numpy.cov``, a ``RuntimeWarning`` for non-positive effective
+    degrees of freedom is only emitted on the unweighted path. The
+    weighted path omits the check so that lazy backends (e.g. Dask) can
+    stay lazy end-to-end; choose ``correction`` and weights such that the
+    effective normalizer is positive.
 
     Examples
     --------
@@ -164,16 +211,57 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     if xp is None:
         xp = array_namespace(m)
 
-    if (
-        is_numpy_namespace(xp)
-        or is_cupy_namespace(xp)
-        or is_torch_namespace(xp)
-        or is_dask_namespace(xp)
-        or is_jax_namespace(xp)
-    ) and m.ndim <= 2:
-        return xp.cov(m)
+    # Validate axis against m.ndim.
+    ndim = max(m.ndim, 1)
+    if not -ndim <= axis < ndim:
+        msg = f"axis {axis} is out of bounds for array of dimension {m.ndim}"
+        raise IndexError(msg)
 
-    return _funcs.cov(m, xp=xp)
+    # Normalize: observations on the last axis. After this, every backend
+    # sees the same convention and we never need to deal with `rowvar`.
+    if m.ndim >= 2 and axis not in (-1, m.ndim - 1):
+        m = xp.moveaxis(m, axis, -1)
+
+    # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
+    # requires integer `correction`. For non-integer-valued `correction`,
+    # fall through to the generic implementation.
+    integer_correction = isinstance(correction, int) or correction.is_integer()
+    has_weights = frequency_weights is not None or weights is not None
+
+    if m.ndim <= 2 and integer_correction:
+        if is_torch_namespace(xp):
+            device = get_device(m)
+            fw = (
+                None
+                if frequency_weights is None
+                else xp.asarray(frequency_weights, device=device)
+            )
+            aw = None if weights is None else xp.asarray(weights, device=device)
+            return xp.cov(m, correction=int(correction), fweights=fw, aweights=aw)
+        # `dask.array.cov` forces `.compute()` whenever weights are given:
+        # its internal `if fact <= 0` check on a lazy 0-D scalar triggers
+        # materialization. Route to the generic impl, which is fully lazy
+        # because it only does sum/matmul and skips that scalar check.
+        if (
+            is_numpy_namespace(xp)
+            or is_cupy_namespace(xp)
+            or is_jax_namespace(xp)
+            or (is_dask_namespace(xp) and not has_weights)
+        ):
+            return xp.cov(
+                m,
+                ddof=int(correction),
+                fweights=frequency_weights,
+                aweights=weights,
+            )
+
+    return _funcs.cov(
+        m,
+        correction=correction,
+        frequency_weights=frequency_weights,
+        weights=weights,
+        xp=xp,
+    )
 
 
 def create_diagonal(
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -230,9 +230,7 @@ def cov(
 
     if m.ndim <= 2 and integer_correction:
         if is_torch_namespace(xp):
-            fw = (
-                None if frequency_weights is None else xp.asarray(frequency_weights)
-            )
+            fw = None if frequency_weights is None else xp.asarray(frequency_weights)
             aw = None if weights is None else xp.asarray(weights)
             return xp.cov(m, correction=int(correction), fweights=fw, aweights=aw)
         # `dask.array.cov` forces `.compute()` whenever weights are given:
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -222,20 +222,6 @@ def cov(
     if m.ndim >= 2 and axis not in (-1, m.ndim - 1):
         m = xp.moveaxis(m, axis, -1)
 
-    # Validate weight shapes (eager metadata, lazy-safe). Value-based
-    # checks (non-negative, integer dtype) are intentionally skipped so
-    # lazy backends don't trigger compute -- same tradeoff as dask.cov.
-    n_obs = m.shape[-1]
-    for name, w in (("fweights", fweights), ("aweights", aweights)):
-        if w is None:
-            continue
-        if w.ndim != 1:
-            msg = f"`{name}` must be 1-D, got ndim={w.ndim}"
-            raise ValueError(msg)
-        if w.shape[0] != n_obs:
-            msg = f"`{name}` has length {w.shape[0]} but `m` has {n_obs} observations"
-            raise ValueError(msg)
-
     # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
     # requires integer `correction`. For non-integer-valued `correction`,
     # fall through to the generic implementation.
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -230,13 +230,10 @@ def cov(
 
     if m.ndim <= 2 and integer_correction:
         if is_torch_namespace(xp):
-            device = get_device(m)
             fw = (
-                None
-                if frequency_weights is None
-                else xp.asarray(frequency_weights, device=device)
+                None if frequency_weights is None else xp.asarray(frequency_weights)
             )
-            aw = None if weights is None else xp.asarray(weights, device=device)
+            aw = None if weights is None else xp.asarray(weights)
             return xp.cov(m, correction=int(correction), fweights=fw, aweights=aw)
         # `dask.array.cov` forces `.compute()` whenever weights are given:
         # its internal `if fact <= 0` check on a lazy 0-D scalar triggers
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -87,8 +87,8 @@ def cov(
     *,
     axis: int = -1,
     correction: int | float = 1,
-    frequency_weights: Array | None = None,
-    weights: Array | None = None,
+    fweights: Array | None = None,
+    aweights: Array | None = None,
     xp: ModuleType | None = None,
 ) -> Array:
     """
@@ -126,12 +126,12 @@ def cov(
         ``correction`` in ``numpy.var``/``std`` and ``torch.cov``.
     fweights : array, optional
         1-D array of integer frequency weights: the number of times each
-        observation is repeated. Corresponds to ``fweights`` in
+        observation is repeated. Same as ``fweights`` in
         ``numpy.cov``/``torch.cov``.
     aweights : array, optional
         1-D array of observation-vector weights (analytic weights). Larger
-        values mark more important observations. Corresponds to
-        ``aweights`` in ``numpy.cov``/``torch.cov``.
+        values mark more important observations. Same as ``aweights`` in
+        ``numpy.cov``/``torch.cov``.
     xp : array_namespace, optional
         The standard-compatible namespace for `m`. Default: infer.
 
@@ -149,8 +149,8 @@ def cov(
         numpy.cov(m, rowvar=False)          -> cov(m, axis=-2)
         numpy.cov(m, bias=True)             -> cov(m, correction=0)
         numpy.cov(m, ddof=k)                -> cov(m, correction=k)
-        numpy.cov(m, fweights=f)            -> cov(m, frequency_weights=f)
-        numpy.cov(m, aweights=a)            -> cov(m, weights=a)
+        numpy.cov(m, fweights=f)            -> cov(m, fweights=f)
+        numpy.cov(m, aweights=a)            -> cov(m, aweights=a)
 
     Unlike ``numpy.cov``, a ``RuntimeWarning`` for non-positive effective
     degrees of freedom is only emitted on the unweighted path. The
@@ -226,12 +226,12 @@ def cov(
     # requires integer `correction`. For non-integer-valued `correction`,
     # fall through to the generic implementation.
     integer_correction = isinstance(correction, int) or correction.is_integer()
-    has_weights = frequency_weights is not None or weights is not None
+    has_weights = fweights is not None or aweights is not None
 
     if m.ndim <= 2 and integer_correction:
         if is_torch_namespace(xp):
-            fw = None if frequency_weights is None else xp.asarray(frequency_weights)
-            aw = None if weights is None else xp.asarray(weights)
+            fw = None if fweights is None else xp.asarray(fweights)
+            aw = None if aweights is None else xp.asarray(aweights)
             return xp.cov(m, correction=int(correction), fweights=fw, aweights=aw)
         # `dask.array.cov` forces `.compute()` whenever weights are given:
         # its internal `if fact <= 0` check on a lazy 0-D scalar triggers
@@ -246,15 +246,15 @@ def cov(
             return xp.cov(
                 m,
                 ddof=int(correction),
-                fweights=frequency_weights,
-                aweights=weights,
+                fweights=fweights,
+                aweights=aweights,
             )
 
     return _funcs.cov(
         m,
         correction=correction,
-        frequency_weights=frequency_weights,
-        weights=weights,
+        fweights=fweights,
+        aweights=aweights,
         xp=xp,
     )
 
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -124,11 +124,11 @@ def cov(
         ``bias=False``). Set to ``0`` for the biased estimate (``N``
         normalization). Corresponds to ``ddof`` in ``numpy.cov`` and to
         ``correction`` in ``numpy.var``/``std`` and ``torch.cov``.
-    frequency_weights : array, optional
+    fweights : array, optional
         1-D array of integer frequency weights: the number of times each
         observation is repeated. Corresponds to ``fweights`` in
         ``numpy.cov``/``torch.cov``.
-    weights : array, optional
+    aweights : array, optional
         1-D array of observation-vector weights (analytic weights). Larger
         values mark more important observations. Corresponds to
         ``aweights`` in ``numpy.cov``/``torch.cov``.
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -222,6 +222,22 @@ def cov(
     if m.ndim >= 2 and axis not in (-1, m.ndim - 1):
         m = xp.moveaxis(m, axis, -1)
 
+    # Validate weight shapes (eager metadata, lazy-safe). Value-based
+    # checks (non-negative, integer dtype) are intentionally skipped so
+    # lazy backends don't trigger compute -- same tradeoff as dask.cov.
+    n_obs = m.shape[-1]
+    for name, w in (("fweights", fweights), ("aweights", aweights)):
+        if w is None:
+            continue
+        if w.ndim != 1:
+            msg = f"`{name}` must be 1-D, got ndim={w.ndim}"
+            raise ValueError(msg)
+        if w.shape[0] != n_obs:
+            msg = (
+                f"`{name}` has length {w.shape[0]} but `m` has {n_obs} observations"
+            )
+            raise ValueError(msg)
+
     # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
     # requires integer `correction`. For non-integer-valued `correction`,
     # fall through to the generic implementation.
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -209,7 +209,7 @@ def cov(
     """
 
     if xp is None:
-        xp = array_namespace(m)
+        xp = array_namespace(m, fweights, aweights)
 
     # Validate axis against m.ndim.
     ndim = max(m.ndim, 1)
@@ -233,9 +233,7 @@ def cov(
             msg = f"`{name}` must be 1-D, got ndim={w.ndim}"
             raise ValueError(msg)
         if w.shape[0] != n_obs:
-            msg = (
-                f"`{name}` has length {w.shape[0]} but `m` has {n_obs} observations"
-            )
+            msg = f"`{name}` has length {w.shape[0]} but `m` has {n_obs} observations"
             raise ValueError(msg)
 
     # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -169,7 +169,16 @@ def broadcast_shapes(
     return _funcs.broadcast_shapes(*shapes)
 
 
-def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
+def cov(
+    m: Array,
+    /,
+    *,
+    axis: int = -1,
+    correction: int | float = 1,
+    frequency_weights: Array | None = None,
+    weights: Array | None = None,
+    xp: ArrayNamespace | None = None,
+) -> Array:
     """
     Estimate a covariance matrix (or a stack of covariance matrices).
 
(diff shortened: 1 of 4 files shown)


betatim:
It looks like the `cov` you are adding follows the pytorch signature, can you explain a bit why you chose that? In my PR I thought following the Numpy API makes sense because it seems that most libraries use that.

The PR description mentions that other functions in this library already use `correction` and `axis`. Which is a good reason to also do it here? Interested in your thinking.

src/array_api_extra/_delegation.py
```diff
@@ -164,16 +211,57 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+
+    # Normalize: observations on the last axis. After this, every backend
+    # sees the same convention and we never need to deal with `rowvar`.
+    if m.ndim >= 2 and axis not in (-1, m.ndim - 1):
+        m = xp.moveaxis(m, axis, -1)
+
+    # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
+    # requires integer `correction`. For non-integer-valued `correction`,
+    # fall through to the generic implementation.
+    integer_correction = isinstance(correction, int) or correction.is_integer()
```
betatim, line 228: Why do we allow non integer corrections in the first place? Is this to allow people to pass `correction=1.` instead of raising an error? Or do people really use corrections that are `correction=1.234` (I'm not familiar with advanced uses)

src/array_api_extra/_delegation.py
```diff
@@ -164,16 +211,57 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    if m.ndim <= 2 and integer_correction:
+        if is_torch_namespace(xp):
+            device = get_device(m)
+            fw = (
+                None
+                if frequency_weights is None
+                else xp.asarray(frequency_weights, device=device)
+            )
+            aw = None if weights is None else xp.asarray(weights, device=device)
+            return xp.cov(m, correction=int(correction), fweights=fw, aweights=aw)
```
betatim, line 235:
```suggestion
            return xp.cov(m, correction=correction, fweights=fw, aweights=aw)
```
shouldn't it already be an `int` if we hit this branch?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,67 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
+    m = xp.asarray(m)
     dtype = (
         xp.float64 if xp.isdtype(m.dtype, "integral") else xp.result_type(m, xp.float64)
     )
 
     m = atleast_nd(m, ndim=2, xp=xp)
     m = xp.astype(m, dtype)
 
-    avg = xp.mean(m, axis=-1, keepdims=True)
+    device = _compat.device(m)
```
betatim: I thought we could assume that all inputs are on the right device already/it is an error to pass arguments that are on different devices?

bruAristimunha:
Hey @betatim!

This was a little hard decision that I had to make, but I can be more strict with numpy if you prefer. 

I basically looked at what was already implemented on the API array and how they handle the parameter names that I was trying to implement. 

Like, for each parameter that I was trying to introduce, I checked how it was made in the past here from numpy to: the `bias`, the `rowvar`, the `ddof`, the `fweights`, and the `aweights`.

Basically, for the `bias`, `ddof` to become `correction`, I notice that in the functions xp.var, xp.std, and think xp.sum, they change the default names to the array api specification name. 

https://data-apis.org/array-api/latest/API_specification/generated/array_api.var.html
https://data-apis.org/array-api/latest/API_specification/generated/array_api.std.html

There was a discussion on how to use correction instead of bias+ddof on these functions. Here was introduced https://github.com/data-apis/array-api/issues/10, and then, later, they made some interesting discussions here: https://github.com/data-apis/array-api/issues/695; it was @kgryte who led the discussion. 

For the case of the `rowvar` becoming the `axis`, I just follow the signature of the other functions. seems like the axis was how they followed. 

And for the frequency_weights and weights, it was my experience in Pyriemann that made the decisions. I think the only place that I remember using something similar was the statsmodels (freq_weights, var_weights) that uses https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.html#statsmodels.genmod.generalized_linear_model.GLM.freq_weights

I think in scikit you guys use sample_weight more, but I can accommodate any request about this.

betatim (commented):
What is your thinking on validating the weights passed in? Things like checking the shapes make sense, that they are all positive (is this actually required? how does it fit with being lazy?)

src/array_api_extra/_delegation.py
bruAristimunha: Here, I follow the xp.var approach to allow int and float, https://data-apis.org/array-api/latest/API_specification/generated/array_api.var.html.

Think about the case use, I don't know either. We can reduce the scope of the var and allow only integers. I think it is a good idea!

bruAristimunha:
I liked this idea a lot @betatim! I think it will make the check in the library that use api array extra much lighter.

src/array_api_extra/_delegation.py
betatim: Maybe @lucascolley has an idea for use-cases?

src/array_api_extra/_lib/_funcs.py
bruAristimunha: very good point! I was thinking that it was necessary to have the device somewhere, if we can assume that it always in the correct device, i think we can remove the whole device reference.

src/array_api_extra/_delegation.py
kgryte: If observations have weights, the unbiased correction is often not `n-1`. Instead, it depends on the sum of weights and their dispersion. Another instance where the correction is not an integer is for autocorrelated data.

One of the reasons for not using `ddof` was to get away from the implicit integer-encoded mental model of correction factors.

src/array_api_extra/_lib/_funcs.py
bruAristimunha: I solved in the last commit.

bruAristimunha:
FYI @qbarthelemy and @agramfort

src/array_api_extra/_delegation.py
bruAristimunha: many thanks for the input @kgryte 🙏🏽

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,62 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
 
     m = atleast_nd(m, ndim=2, xp=xp)
     m = xp.astype(m, dtype)
 
-    avg = xp.mean(m, axis=-1, keepdims=True)
+    fw = (
+        None
+        if frequency_weights is None
+        else xp.astype(xp.asarray(frequency_weights), dtype)
+    )
```
qbarthelemy:
```suggestion
    fw = None
    if frequency_weights is not None:
        fw = xp.astype(xp.asarray(frequency_weights), dtype)
```

src/array_api_extra/_delegation.py
```diff
@@ -92,16 +101,37 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        ``N - correction`` (for unweighted input). Default: ``1``, which
+        gives the unbiased estimate (matches ``numpy.cov`` default of
+        ``bias=False``). Set to ``0`` for the biased estimate (``N``
+        normalization). Corresponds to ``ddof`` in ``numpy.cov`` and to
+        ``correction`` in ``numpy.var``/``std`` and ``torch.cov``.
+    frequency_weights : array, optional
+        1-D array of integer frequency weights: the number of times each
+        observation is repeated. Corresponds to ``fweights`` in
+        ``numpy.cov``/``torch.cov``.
+    weights : array, optional
```
qbarthelemy: My advice would be to use the argument names from NumPy: `bias`, `ddof`, `fweights`, `aweights`.
As it is, it's a rather unreadable mix between NumPy and PyTorch.
```suggestion
    fweights : array, optional
        1-D array of integer frequency weights: the number of times each
        observation is repeated. Corresponds to ``fweights`` in
        ``numpy.cov``/``torch.cov``.
    aweights : array, optional
```

bruAristimunha: I think bias and ddof is not possible because of the discussion above. But yes for the weights parameters

bruAristimunha: https://github.com/data-apis/array-api-extra/pull/690#issuecomment-4279522152

qbarthelemy: I updated myself by reading the link, okay for `correction`

betatim:
Thanks a lot for the detailed answer in https://github.com/data-apis/array-api-extra/pull/690#issuecomment-4279522152 - I didn't realise there was precedent for using `correction` in functions like `var`. I think it makes sense to copy that and use `correction` for `cov` as well. Worth making the translation!

What is the "temporary deployed" thing that keeps happening?

bruAristimunha:
it is not me @betatim, i think it something that @lucascolley is pushing in pushing here: https://github.com/data-apis/array-api-extra/pull/699

bruAristimunha:
Happy that you liked the response @betatim :)

I think I addressed all the points from you and @qbarthelemy, can we merge?

lucascolley:
> What is the "temporary deployed" thing that keeps happening?

fixed in https://github.com/data-apis/array-api-extra/commit/bd3652ad77aa1f9a382492d771071f6689128a75

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,62 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
+    /,
+    *,
+    correction: int | float = 1,
+    fweights: Array | None = None,
+    aweights: Array | None = None,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
     """See docstring in array_api_extra._delegation."""
-    m = xp.asarray(m, copy=True)
+    m = xp.asarray(m)
```
lucascolley, line 294: why do we drop this?

src/array_api_extra/_delegation.py
```diff
@@ -164,16 +211,68 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     if xp is None:
         xp = array_namespace(m)
```
lucascolley:
```suggestion
        xp = array_namespace(m, fweights, aweights)
```

src/array_api_extra/_delegation.py
```diff
@@ -164,16 +211,68 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
-        or is_jax_namespace(xp)
-    ) and m.ndim <= 2:
-        return xp.cov(m)
-
-    return _funcs.cov(m, xp=xp)
+    # Validate axis against m.ndim.
+    ndim = max(m.ndim, 1)
+    if not -ndim <= axis < ndim:
+        msg = f"axis {axis} is out of bounds for array of dimension {m.ndim}"
+        raise IndexError(msg)
```
lucascolley, line 218: just a thought, maybe some common logic can be extracted out from this and https://github.com/data-apis/array-api-extra/blob/af12cd5653002bf7160cd14ea461d594971e5966/src/array_api_extra/_delegation.py#L313-L316

that is for a tuple of axes though, so maybe not

src/array_api_extra/_delegation.py
```diff
@@ -164,16 +211,68 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        if w is None:
+            continue
+        if w.ndim != 1:
+            msg = f"`{name}` must be 1-D, got ndim={w.ndim}"
+            raise ValueError(msg)
+        if w.shape[0] != n_obs:
+            msg = (
+                f"`{name}` has length {w.shape[0]} but `m` has {n_obs} observations"
+            )
+            raise ValueError(msg)
```
lucascolley: must this happen at the delegation layer? What happens if we just let the backends to which we delegate error out instead?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,62 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
+        v1 = xp.sum(w, axis=-1)
+        avg = xp.sum(m * w, axis=-1, keepdims=True) / v1
+        if aw is None:
+            fact = v1 - correction
+        else:
+            fact = v1 - correction * xp.sum(w * aw, axis=-1) / v1
+
+    m_c = m - avg
+    m_w = m_c if w is None else m_c * w
+    m_cT = xp.matrix_transpose(m_c)
```
lucascolley, line 353: minor nit: can we just use `m_c.mT` instead?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,62 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
+    if xp.isdtype(m_cT.dtype, "complex floating"):
+        m_cT = xp.conj(m_cT)
+    c = xp.matmul(m_w, m_cT) / fact
```
lucascolley: minor nit: can we use `m_w @ m_cT` instead?

lucascolley (commented):
I took an initial look, seems pretty good!

One high level comment @bruAristimunha — could you demonstrate that this works as expected when used in a branch of sklearn? You should be able to change https://github.com/scikit-learn/scikit-learn/blob/06aded051fe6c7c9970b7e13c3669f952a799831/maint_tools/vendor_array_api_extra.sh#L8-L9 to point to this branch and commit hash.

src/array_api_extra/_lib/_funcs.py
bruAristimunha: hey @lucascolley! I tried in the first iteration 😢, but Dask doesn't have the .mT, so fail the whole test suite.

One alternative that I tested was to try to use Dask directly to compute the covariance, and this will fail the whole test suite for the lazy behavior.

The root of the problem is this line on Dask:
https://github.com/dask/dask/blob/2026.3.0/dask/array/routines.py#L1605

lucascolley: ohh yeah, feel free to ~either~ leave this as is

bruAristimunha: I can put as non-lazy just for dask, but I think it is okay to keep the transpose function.

src/array_api_extra/_delegation.py
bruAristimunha: I think all the back-end (numpy, torch, jax, dask), except the 0-D scalar fweights for dask, will raise, and should be fine.

I mostly push this to address the suggestion of @betatim (https://github.com/data-apis/array-api-extra/pull/690#pullrequestreview-4138954870), but if you prefer, I can let the backend handle this, as they already validate.

bruAristimunha: let's me validate only in the case of dask

bruAristimunha: moving this for the _funcs.

bruAristimunha: e34d415ec22eaa3f069fa0d3a35b62ed1d4e181d

src/array_api_extra/_delegation.py
bruAristimunha: I couldn't think of something to optimize this. The only thing that i could think here is the normalize_axis_index from numpy, but then we would need another PR to introduce here in the library ;p

https://numpy.org/doc/2.1/reference/generated/numpy.lib.array_utils.normalize_axis_index.html#numpy.lib.array_utils.normalize_axis_index

src/array_api_extra/_lib/_funcs.py
bruAristimunha: It was mostly one small optimization that I made, as the new code for the covariance doesn't mutate in-place anymore.

Like, if I understand correctly, we need to copy because of this line:

```
 m -= avg 
```

But now, as we are doing:

```python
m_c = m - avg                                                                                                                               
m_w = m_c if w is None else m_c * w                                                                                                                                                        
m_cT = xp.matrix_transpose(m_c)
c = (m_w @ m_cT) / fact
```

I noticed this by accident, then I was testing the speed test, and I noticed some small regression. I think it's worth disabling copying.

bruAristimunha:
hey @betatim, 

As you have the first covariance PR on scikit, can you help with this small test as requested by @lucascolley?

> 
> One high level comment @bruAristimunha — could you demonstrate that this works as expected when used in a branch of sklearn? You should be able to change https://github.com/scikit-learn/scikit-learn/blob/06aded051fe6c7c9970b7e13c3669f952a799831/maint_tools/vendor_array_api_extra.sh#L8-L9 to point to this branch and commit hash.

bruAristimunha:
hey @lucascolley,

I made in my branch that was built on top of @betatim's work for scikit-learn first covariance, you can check more here: https://github.com/scikit-learn/scikit-learn/pull/33600

lucascolley:
> hey @lucascolley,
> 
> I made in my branch that was built on top of @betatim's work for scikit-learn first covariance, you can check more here: [scikit-learn/scikit-learn#33600](https://github.com/scikit-learn/scikit-learn/pull/33600)

thanks! Would be great if you could take a look, Tim

src/array_api_extra/_lib/_funcs.py
```diff
@@ -281,31 +281,79 @@ def broadcast_shapes(*shapes: tuple[float | None, ...]) -> tuple[int | None, ...
...
+            fact = v1 - correction
+        else:
+            fact = v1 - correction * xp.sum(w * aw, axis=-1) / v1
+
+    m_c = m - avg
+    m_w = m_c if w is None else m_c * w
+    m_cT = xp.matrix_transpose(m_c)
+    if xp.isdtype(m_cT.dtype, "complex floating"):
+        m_cT = xp.conj(m_cT)
+    c = (m_w @ m_cT) / fact
```
qbarthelemy:
```suggestion
    c = m_w @ m_cT / fact
```

bruAristimunha:
hey @lucascolley,

I was wondering, can you please approve the CI for the final test?

bruAristimunha:
hey @lucascolley,

All good now! Let's merge? :)

lucascolley:
@betatim did you get a chance to take a look at https://github.com/scikit-learn/scikit-learn/pull/33600 ?

betatim:
That PR has a huge diff so I'm not sure what to look for.

I think in scikit-learn we are leaning towards using the "post-hoc centering" approach (eg what `PCA` does with the `covariance_eigh` solver). It avoids having to copy `X` which is a big win in terms of memory usage. https://github.com/scikit-learn/scikit-learn/pull/33573#discussion_r3101245545

The thing that I don't have straight in my mind yet is the answer to the question: "why doesn't everyone do this all the time?". I assume there is a reason, because it isn't like post-hoc centering was just invented.

---

I think in this PR/`array-api-extra` we should implement `cov` in the way that numpy does. My hunch is that if we make it use the post-hoc centering approach we might catch people off-guard?!

bruAristimunha:
heeey @betatim!

I agree that I may have another ongoing discussion about the covariance with pos-hoc, and that some decisions will need to be made in scikit-learn.

But I think in this PR we are mostly exposing the parameters we don't have at the moment in the api array extra, using the native implementation, nothing more. I understand that the discussion about whether to include the parameter for post-hoc analysis needs to go through numpy first. 

Regarding the long PR, I can make it more compact and more comprehensive. I feel some details of how to handle the parameters, etc., may not be super clear.

bruAristimunha:
Like, the pos-hoc is not implemented here.

lucascolley:
> That PR has a huge diff so I'm not sure what to look for.

The main thing I'm interested in is 'does the interface implemented here appear to work as intended when used in sklearn?'

betatim:
Quick follow up: the PR I was referring to as "huge diff" is https://github.com/scikit-learn/scikit-learn/pull/33600

I think this PR (#690) is fine.

---

With the discussion in scikit-learn about using `np.cov` style covariance calculation vs using post-hoc centering it means we might not actually end up using `xpx.cov` (in the current PRs). So, in isolation/without a concrete use-case in mind I think we should try and follow the array API conventions on naming of paramters and such (as this PR does).

bruAristimunha:
Hey @lucascolley,

After many conversations, scikit-learn decided to reimplement some post-hoc recentering, which is still not available in numpy or torch.

There is some ongoing discussion to implement this in numpy, but it is likely to take some time.

So, the request to validate in scikit will not be possible at the moment, even though I show the use in my commit branch.

Let us merge this PR and consider this context. Covariance matrices are widely used in the machine learning community, and the normalization of parameters I propose is important, IHMO.

I am happy to discuss and iterate on any further requests and appreciate your time dedicated to this!

bruAristimunha:
The discussion is going here: https://github.com/numpy/numpy/issues/31292

lucascolley:
> So, the request to validate in scikit will not be possible at the moment, even though I show the use in my commit branch.

maybe there is a different library we can check with?

bruAristimunha:
I can check with pyriemann @lucascolle! cc. @qbarthelemy, @agramfort

lucascolley:
gentle bump :)

bruAristimunha:
hey @lucascolley,

Thanks for pushing this test. I captured some errors that were breaking the gradient. Now everything is working, can you please check the pr at Pyriemann and merge this one? 🙏🏽

src/array_api_extra/_lib/_funcs.py
```diff
@@ -259,31 +246,81 @@ def broadcast_shapes(  # numpydoc ignore=PR01,RT01
...
+    correction: int | float = 1,
+    fweights: Array | None = None,
+    aweights: Array | None = None,
+    xp: ModuleType,
+) -> Array:  # numpydoc ignore=PR01,RT01
     """See docstring in array_api_extra._delegation."""
-    m = xp.asarray(m, copy=True)
+    # NB: no `xp.asarray(m)` here. The delegation layer already guarantees `m`
+    # is an array (it calls `array_namespace(m)` and reads `m.ndim`), and on
+    # torch `xp.asarray` detaches gradients and mutates the caller's tensor.
```
lucascolley, line 278: Yes I am comfortable to remove `asarray`, but are we definitely fine to remove the `copy`? Is a copy guaranteed elsewhere?

lucascolley (commented):
great, thanks for following up!

Could you take a look at that PR perhaps @qbarthelemy ?

qbarthelemy (approved):
It's ok for me. Good job @bruAristimunha !

lucascolley:
I will try to review this soon then take a look at making a release

src/array_api_extra/_delegation.py
```diff
@@ -177,16 +163,44 @@ def cov(m: Array, /, *, xp: ModuleType | None = None) -> Array:
     :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
     of :math:`x_i`.
 
-    With the exception of supporting batch input, this provides a subset of
-    the functionality of ``numpy.cov``.
+    Extends ``numpy.cov`` with support for batch input and array-api
+    backends. Naming follows the array-api conventions used elsewhere in
+    this library (``axis``, ``correction``) rather than the numpy spellings
+    (``rowvar``, ``bias``, ``ddof``); see Notes for the mapping.
```
lucascolley:
```suggestion
    Extends ``numpy.cov`` with support for batch input.
    Naming follows the array API conventions used elsewhere in
    this library (``axis``, ``correction``) rather than the NumPy spellings
    (``rowvar``, ``bias``, ``ddof``); see Notes for the mapping.
```

src/array_api_extra/_lib/_funcs.py
```diff
@@ -260,31 +260,81 @@ def broadcast_shapes(  # numpydoc ignore=PR01,RT01
...
+    # sparse, and the dask+weights fallback where the native check is
+    # bypassed to preserve laziness).
+    n_obs = m.shape[-1]
+    for name, w_in in (("fweights", fweights), ("aweights", aweights)):
+        if w_in is None:
+            continue
+        if w_in.ndim != 1:
+            msg = f"`{name}` must be 1-D, got ndim={w_in.ndim}"
+            raise ValueError(msg)
+        if w_in.shape[0] != n_obs:
```
lucascolley: what happens here if `n_obs` or `w_in` have a lazy shape? Do we want to avoid this error when `w_in.shape[0] == n_obs == m.shape[-1] == None`?

src/array_api_extra/_delegation.py
```diff
@@ -196,6 +233,23 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+        numpy.cov(m, bias=True)             -> cov(m, correction=0)
+        numpy.cov(m, ddof=k)                -> cov(m, correction=k)
+        numpy.cov(m, fweights=f)            -> cov(m, fweights=f)
+        numpy.cov(m, aweights=a)            -> cov(m, aweights=a)
+
+    Unlike ``numpy.cov``, a ``RuntimeWarning`` for non-positive effective
+    degrees of freedom is only emitted on the unweighted path. The
+    weighted path omits the check so that lazy backends (e.g. Dask) can
+    stay lazy end-to-end; choose ``correction`` and weights such that the
+    effective normalizer is positive.
```
lucascolley: hmm, can't we instead keep warning on the weighted path, but only when https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_lazy_array is false? That is what we do at e.g. https://github.com/scipy/scipy/blob/3f71ae04a8e02a44687f9557e7c12be60b5e1160/scipy/cluster/hierarchy/_hierarchy_impl.py#L941-L947.

lucascolley (commented):
a few more comments, sorry for not a full review yet

bruAristimunha:
should be fine now @lucascolley

src/array_api_extra/_lib/_funcs.py
```diff
@@ -263,31 +263,102 @@ def broadcast_shapes(  # numpydoc ignore=PR01,RT01
...
-
-    m_shape = eager_shape(m)
-    fact = m_shape[-1] - 1
+    # Preserve the historical no-alias guarantee even when the dtype already matches.
+    m = xp.astype(m, dtype, copy=True)
+
+    # Validate weight shapes (eager metadata, lazy-safe). Native backends
+    # validate themselves; this covers the generic path (array-api-strict,
+    # sparse, and the dask+weights fallback where the native check is
+    # bypassed to preserve laziness).
```
lucascolley: I think this is stating the obvious

```suggestion
    # Validate weight shapes (eager metadata, lazy-safe).
```

src/array_api_extra/_delegation.py
```diff
@@ -249,18 +303,54 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+
+    # Normalize: observations on the last axis. After this, every backend
+    # sees the same convention and we never need to deal with `rowvar`.
+    if m.ndim >= 2 and axis not in (-1, m.ndim - 1):
+        m = xp.moveaxis(m, axis, -1)
+
+    # `numpy.cov` (and cupy/dask/jax) require integer `ddof`; `torch.cov`
+    # requires integer `correction`. For non-integer-valued `correction`,
+    # fall through to the generic implementation.
+    integer_correction = isinstance(correction, int) or correction.is_integer()
```
lucascolley: what is this `.is_integer()` method?

src/array_api_extra/_delegation.py
lucascolley: this comment was resolved, but I think it still applies?

If this is included for the static type checkers, we can use `typing.cast` instead, I think.

src/array_api_extra/_delegation.py
```diff
@@ -249,18 +303,54 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+        # because it only does sum/matmul and skips that scalar check.
+        if (
+            is_numpy_namespace(xp)
+            or is_cupy_namespace(xp)
+            or is_jax_namespace(xp)
+            or (is_dask_namespace(xp) and not has_weights)
+        ):
+            return xp.cov(
+                m,
+                ddof=int(correction),
```
lucascolley: again, why do we need the `int` cast?

src/array_api_extra/_delegation.py
```diff
@@ -180,16 +189,44 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
     :math:`x_i` and :math:`x_j`. The element :math:`C_{ii}` is the variance
     of :math:`x_i`.
 
-    With the exception of supporting batch input, this provides a subset of
-    the functionality of ``numpy.cov``.
+    Extends ``numpy.cov`` with support for batch input.
```
lucascolley:
```suggestion
    Extends :func:`numpy.cov` with support for batch input.
```

src/array_api_extra/_delegation.py
lucascolley:
```suggestion
    :math:`N \\times N` covariance matrix is the covariance of
```

src/array_api_extra/_delegation.py
```diff
@@ -180,16 +189,44 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
     m : array
         An array of shape ``(..., N, M)`` whose innermost two dimensions
-        contain *M* observations of *N* variables. That is,
-        each row of `m` represents a variable, and each column a single
-        observation of all those variables.
+        contain *M* observations of *N* variables by default. The axis of
+        observations is controlled by `axis`.
+    axis : int, optional
+        Axis of `m` containing the observations. Default: ``-1`` (the last
+        axis), matching the array-api convention. Use ``axis=-2`` (or ``0``
```
lucascolley:
```suggestion
        axis), matching the array API convention. Use ``axis=-2`` (or ``0``
```

src/array_api_extra/_delegation.py
```diff
@@ -180,16 +189,44 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+        for 2-D input) to treat each column as a variable, which
+        corresponds to ``rowvar=False`` in ``numpy.cov``.
```
lucascolley:
```suggestion
        corresponds to ``rowvar=False`` in :func:`numpy.cov`.
```

src/array_api_extra/_delegation.py
```diff
@@ -180,16 +189,44 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+        ``bias=False``). Set to ``0`` for the biased estimate (``N``
+        normalization). Corresponds to ``ddof`` in ``numpy.cov`` and to
+        ``correction`` in ``numpy.var``/``std`` and ``torch.cov``.
+        Non-integer values are allowed for advanced use cases: the
+        unbiased correction for weighted observations depends on the
+        sum and dispersion of the weights and is generally not an
+        integer, and autocorrelated data may also require a fractional
+        correction. Non-integer ``correction`` routes through the
+        generic implementation because ``numpy.cov``'s ``ddof`` and
+        ``torch.cov``'s ``correction`` both require integers.
```
lucascolley: Would be good to use intersphinx `:func:` links for the rest of the references to `numpy` and `torch` functions that appear in the docstring.

src/array_api_extra/_delegation.py
```diff
@@ -199,6 +236,23 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
         An array having shape (..., N, N) whose innermost two dimensions represent
```
lucascolley:
```suggestion
        An array having shape ``(..., N, N)`` whose innermost two dimensions represent
```

src/array_api_extra/_delegation.py
```diff
@@ -199,6 +236,23 @@ def cov(m: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
...
+        numpy.cov(m, fweights=f)            -> cov(m, fweights=f)
+        numpy.cov(m, aweights=a)            -> cov(m, aweights=a)
+
+    A ``RuntimeWarning`` is emitted for non-positive effective degrees of
+    freedom when the effective normalizer can be checked without materializing
+    a lazy array. When the normalizer itself is lazy (e.g. for weighted Dask
+    inputs), this check is skipped; choose ``correction`` and weights such that
+    it is positive.
+
     Examples
```
lucascolley, line 257: Please can we add to the examples to demonstrate usage of the new parameters?

src/array_api_extra/_lib/_funcs.py
```diff
@@ -263,31 +263,102 @@ def broadcast_shapes(  # numpydoc ignore=PR01,RT01
...
+            continue
+        if w_in.ndim != 1:
+            msg = f"`{name}` must be 1-D, got ndim={w_in.ndim}"
+            raise ValueError(msg)
+        weight_length = w_in.shape[0]
+        if (
+            weight_length is not None
+            and n_obs is not None
+            and not math.isnan(weight_length)
+            and not math.isnan(n_obs)
```
lucascolley, line 302: worth a comment that this accounts for the non-standard behaviour of Dask, using NaN instead of `None` for unknown dimensions

lucascolley (commented):
thanks again for your patience!

I may cut a release before we finish this PR, but hopefully it can be included in the next one :)

src/array_api_extra/_delegation.py
bruAristimunha: It is needed at runtime, not just for the type checkers: this branch also accepts integer-valued floats like `1.0` (`float(correction).is_integer()`), and `torch.cov` rejects those with `TypeError: cov(): argument 'correction' must be int, not float` — so `typing.cast` wouldn't be enough. Hoisted to a single `int_correction = int(correction)` with a comment explaining this in a34ba84.

src/array_api_extra/_delegation.py
bruAristimunha: It's the builtin `float.is_integer()` — the `isinstance` guard was only there because `int.is_integer()` doesn't exist before Python 3.12. Simplified to `float(correction).is_integer()` in a34ba84 to make that obvious.

src/array_api_extra/_delegation.py
bruAristimunha: numpy does accept integer-valued floats for `ddof` at runtime, but its stubs type it as `int`, and the torch branch above needs a real `int` anyway — so both branches now share a single `int_correction = int(correction)`, with a comment, in a34ba84.

src/array_api_extra/_delegation.py
bruAristimunha: Done in a34ba84 — this also needed `torch` added to the intersphinx mapping in `docs/conf.py`.

src/array_api_extra/_delegation.py
bruAristimunha: Added in a34ba84: `correction=0` vs the default on a small variance example, `fweights` shown to be equivalent to repeating observations, and `aweights` down-weighting the extremes.

src/array_api_extra/_lib/_funcs.py
bruAristimunha: Comment added in a34ba84.

bruAristimunha:
Everything is solved @lucascolley, please let me know if any more iterations are necessary

lucascolley (approved):
thanks @bruAristimunha for the perseverance, I think this is ready to merge now — nice work!

Thanks also @qbarthelemy @betatim for the reviews!

BTW I added pyRiemann to the list of users in the README, hope you don't mind 😉 feel free to shout if you hit any problems, or have any other functions that you would like to see added.

lucascolley:
@all-contributors please add @bruAristimunha for examples

lucascolley:
@all-contributors please add @betatim for review

PR #695, opened 2026-04-18
TYP: add `pyrefly`

### LINKED ISSUE #694: TYP: adopt `pyrefly`
Would be cool to see if this is feasible.

lucascolley:
@jorenham could I possibly interest you to review?

closes gh-694

### DIFF
--- a/src/array_api_extra/_lib/_lazy.py
+++ b/src/array_api_extra/_lib/_lazy.py
@@ -30,7 +30,7 @@ else:
 P = ParamSpec("P")
 
 
-@overload
+@overload  # pyrefly: ignore[invalid-param-spec]
 def lazy_apply(  # type: ignore[valid-type]
     func: Callable[P, Array | ArrayLike],
     *args: Array | complex | None,
@@ -42,7 +42,7 @@ def lazy_apply(  # type: ignore[valid-type]
 ) -> Array: ...  # numpydoc ignore=GL08
 
 
-@overload
+@overload  # pyrefly: ignore[invalid-param-spec]
 def lazy_apply(  # type: ignore[valid-type]
     func: Callable[P, Sequence[Array | ArrayLike]],
     *args: Array | complex | None,
@@ -54,7 +54,7 @@ def lazy_apply(  # type: ignore[valid-type]
 ) -> tuple[Array, ...]: ...  # numpydoc ignore=GL08
 
 
-def lazy_apply(  # type: ignore[valid-type]  # numpydoc ignore=GL07,SA04
+def lazy_apply(  # type: ignore[valid-type]  # pyrefly: ignore[invalid-param-spec]  # numpydoc ignore=GL07,SA04
     func: Callable[P, Array | ArrayLike | Sequence[Array | ArrayLike]],
     *args: Array | complex | None,
     shape: tuple[int | None, ...] | Sequence[tuple[int | None, ...]] | None = None,
@@ -240,7 +240,7 @@ def lazy_apply(  # type: ignore[valid-type]  # numpydoc ignore=GL07,SA04
     if is_dask_namespace(xp):
         import dask
 
-        metas: list[Array] = [arg._meta for arg in array_args]  # pylint: disable=protected-access    # pyright: ignore[reportAttributeAccessIssue]
+        metas: list[Array] = [arg._meta for arg in array_args]  # pylint: disable=protected-access # pyright: ignore[reportAttributeAccessIssue] # pyrefly: ignore[missing-attribute]
         meta_xp = array_namespace(*metas)
 
         wrapped = dask.delayed(  # type: ignore[attr-defined]  # pyright: ignore[reportPrivateImportUsage]
--- a/tests/test_lazy.py
+++ b/tests/test_lazy.py
@@ -219,7 +219,10 @@ def test_lazy_apply_none_shape_in_args(xp: ModuleType, library: Backend):
     mxp = np if library is Backend.DASK else xp
     int_type = xp.asarray(0).dtype
 
-    ctx: contextlib.AbstractContextManager[object]
+    ctx: (
+        contextlib.AbstractContextManager[object]
+        | contextlib.AbstractContextManager[None]
+    )
     if library.like(Backend.JAX):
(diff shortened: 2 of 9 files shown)


src/array_api_extra/_lib/_lazy.py
```diff
@@ -30,7 +30,7 @@
 P = ParamSpec("P")
 
 
-@overload
+@overload  # pyrefly: ignore[invalid-param-spec]
```
lucascolley, line 33: pyrefly does not like us using `P.kwargs` without also using `P.args`.

tests/test_lazy.py
```diff
@@ -219,7 +219,10 @@ def test_lazy_apply_none_shape_in_args(xp: ModuleType, library: Backend):
     mxp = np if library is Backend.DASK else xp
     int_type = xp.asarray(0).dtype
 
-    ctx: contextlib.AbstractContextManager[object]
+    ctx: (
+        contextlib.AbstractContextManager[object]
+        | contextlib.AbstractContextManager[None]
+    )
```
lucascolley, line 225: pyrefly is not happy for `contextlib.nullcontext()` to fall under `contextlib.AbstractContextManager[object]`

vendor_tests/_array_api_compat_vendor.py
```diff
@@ -1,11 +1,14 @@
...
+from .array_api_compat import *  # noqa: F403 # pyright: ignore[reportAssignmentType]
 from .array_api_compat import array_namespace as array_namespace_compat
 
 
 # Let unit tests check with `is` that we are picking up the function from this module
 # and not from the original array_api_compat module.
-def array_namespace(*xs, **kwargs):  # numpydoc ignore=GL08
-    return array_namespace_compat(*xs, **kwargs)
+def array_namespace(*xs: Any | complex | None, **kwargs) -> ModuleType:  # type: ignore[no-redef] # numpydoc ignore=GL08
+    return array_namespace_compat(*xs, **kwargs)  # pyright: ignore[reportUnknownArgumentType]
```
lucascolley: this seems like an improvement, but we still don't have a full signature since we want to avoid having to make changes when new `kwargs` are added to `array_api_compat.array_namespace`

vendor_tests/test_vendor.py
lucascolley, line 1: `mypy` and `pyright` were previously not checking the `vendor_tests` dir, that is fixed now

vendor_tests/test_vendor.py
```diff
@@ -35,6 +37,7 @@ def test_vendor_compat():
     to_device(x, device(x))
     assert is_array_api_obj(x)
     assert is_array_api_strict_namespace(xp)
+    x = cast(Array, x)
```
lucascolley, line 42: otherwise `x` stays narrowed by `TypeGuard[_ArrayApiObj]`, where:

```
Argument type is partially unknown
  Argument corresponds to parameter "x" in function "is_cupy_array"
  Argument type is "ndarray[tuple[Any, ...], dtype[Any]] | Unknown | Array | SupportsArrayNamespace[Any]
```

vendor_tests/test_vendor.py
```diff
@@ -1,10 +1,14 @@
 # pyright: reportAttributeAccessIssue=false
 
-from typing import Any
+from typing import Any, cast
 
 import array_api_strict as xp
 from numpy.testing import assert_array_equal
 
+from vendor_tests.array_api_compat.common._typing import (  # type: ignore[import-not-found]
```
lucascolley, line 8: maybe we should be cloning the vendored libraries into place for the static analysis — not sure

src/array_api_extra/_lib/_lazy.py
jorenham: I'm afraid that this is more of a typing spec thing than a pyrefly thing: https://typing.python.org/en/latest/spec/generics.html#id5

If you only have `*args` then you could've used a `TypeVarTuple` (https://typing.python.org/en/latest/spec/generics.html#args-as-a-type-variable-tuple), but for this `**kwargs` case there's no good solution :(

tests/test_lazy.py
jorenham: Sounds like a bug in Pyrefly, because the first typar of `AbstractContextManager` is supposed to be covariant: https://github.com/python/typeshed/blob/eec9fe9aa9ce87df8987de5c1401f743e179378a/stdlib/contextlib.pyi#L48

Do you want to report it or should I?

src/array_api_extra/_lib/_lazy.py
```diff
@@ -240,7 +240,7 @@ def lazy_apply(  # type: ignore[valid-type]  # numpydoc ignore=GL07,SA04
     if is_dask_namespace(xp):
         import dask
 
-        metas: list[Array] = [arg._meta for arg in array_args]  # pylint: disable=protected-access    # pyright: ignore[reportAttributeAccessIssue]
+        metas: list[Array] = [arg._meta for arg in array_args]  # pylint: disable=protected-access # pyright: ignore[reportAttributeAccessIssue] # pyrefly: ignore[missing-attribute]
```
jorenham, line 243: If you want to avoid long lines, then this should also do the trick I believe:

```suggestion
        # pyrefly: ignore[missing-attribute]
        metas: list[Array] = [arg._meta for arg in array_args]  # pylint: disable=protected-access # pyright: ignore[reportAttributeAccessIssue]
```

tests/test_lazy.py
lucascolley: Go for it if you're happy to!

src/array_api_extra/_lib/_lazy.py
lucascolley: could do, but tbh I don't mind long lines as long as they are just comments to the machine

lucascolley (commented):
I guess I'll merge this if it seems alright @jorenham ?

jorenham (approved):

lucascolley:
thanks Joren

tests/test_lazy.py
jorenham: https://github.com/facebook/pyrefly/issues/3046#issuecomment-4279001469

PR #713, opened 2026-05-04
ENH: delegate `broadcast_shapes`

Cyril-36:
Part of #100.\n\nSummary:\n- Move the public broadcast_shapes wrapper into the delegation layer.\n- Delegate known integer shapes to numpy.broadcast_shapes when NumPy is available.\n- Preserve the existing fallback for None, math.nan, and no-NumPy runtimes.\n\nTesting:\n- pixi run -e tests pytest -q tests/test_funcs.py::TestBroadcastShapes\n- pixi run -e tests tests\n- pixi run -e lint ruff check src/array_api_extra/_delegation.py src/array_api_extra/_lib/_funcs.py src/array_api_extra/__init__.py tests/test_funcs.py\n- pixi run -e lint ruff format --check src/array_api_extra/_delegation.py src/array_api_extra/_lib/_funcs.py src/array_api_extra/__init__.py tests/test_funcs.py\n- pixi run -e lint mypy src/array_api_extra/_delegation.py src/array_api_extra/_lib/_funcs.py tests/test_funcs.py\n- pixi run -e lint pyright src/array_api_extra/_delegation.py src/array_api_extra/_lib/_funcs.py tests/test_funcs.py\n- pixi run -e lint pyrefly check src/array_api_extra/_delegation.py src/array_api_extra/_lib/_funcs.py tests/test_funcs.py\n- pixi run -e docs docs

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -2,7 +2,7 @@
 
 from collections.abc import Sequence
 from types import ModuleType
-from typing import Literal
+from typing import Literal, cast
 
 from ._lib import _funcs
 from ._lib._utils._compat import (
@@ -20,6 +20,7 @@ from ._lib._utils._typing import Array, DType
 
 __all__ = [
     "atleast_nd",
+    "broadcast_shapes",
     "cov",
     "create_diagonal",
     "expand_dims",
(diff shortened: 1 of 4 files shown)


src/array_api_extra/_delegation.py
```diff
@@ -81,6 +87,55 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
...
+    >>> xpx.broadcast_shapes((2, 3), (2, 1))
+    (2, 3)
+    >>> xpx.broadcast_shapes((4, 2, 3), (2, 1), (1, 3))
+    (4, 2, 3)
+    """
+    if np is not None and all(
+        isinstance(size, int) for shape in shapes for size in shape
+    ):
+        int_shapes = cast(tuple[tuple[int, ...], ...], shapes)
+        return cast(tuple[int | None, ...], np.broadcast_shapes(*int_shapes))
```
lucascolley: please see how the other functions in this file work — we want to grab `xp`

https://github.com/data-apis/array-api-extra/blob/bc126fa1d46f74f5d5660bc0680a9261ecd813ae/src/array_api_extra/_delegation.py#L220-L221

and then use functions from that namespace where possible:

https://github.com/data-apis/array-api-extra/blob/bc126fa1d46f74f5d5660bc0680a9261ecd813ae/src/array_api_extra/_delegation.py#L227-L236

Cyril-36:
Updated based on Copilot review:\n- Added the numpydoc ignore marker to the internal fallback implementation.\n- Moved the optional NumPy import into a helper so NumPy is not imported at module import time.

lucascolley (changes requested):

Cyril-36:
Thanks for the review. Updated the implementation to follow the existing delegation pattern: `broadcast_shapes` now accepts `xp`, delegates through `xp.broadcast_shapes` for supported namespaces and known integer shapes, and falls back to the generic implementation for unknown sizes or when no namespace is provided.

src/array_api_extra/_delegation.py
```diff
@@ -81,6 +82,68 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
...
+    >>> xpx.broadcast_shapes((4, 2, 3), (2, 1), (1, 3))
+    (4, 2, 3)
+    """
+    if (
+        xp is not None
+        and all(isinstance(size, int) for shape in shapes for size in shape)
+        and (
+            is_numpy_namespace(xp)
+            or is_cupy_namespace(xp)
+            or is_dask_namespace(xp)
```
lucascolley: (dask doesn't have this function)

```suggestion
```

Cyril-36: Thanks, fixed. Removed Dask from the native `broadcast_shapes` delegation path since Dask does not provide this function, so Dask now falls back to the generic implementation.

Cyril-36:
Hi @lucascolley, just following up. I pushed the requested updates: `broadcast_shapes` now accepts `xp`, Dask was removed from the native delegation path, and the previous review comments appear outdated. Please let me know if anything else needs to be changed.

src/array_api_extra/_delegation.py
```diff
@@ -81,6 +82,67 @@ def atleast_nd(x: Array, /, *, ndim: int, xp: ModuleType | None = None) -> Array
...
+    --------
+    >>> import array_api_extra as xpx
+    >>> xpx.broadcast_shapes((2, 3), (2, 1))
+    (2, 3)
+    >>> xpx.broadcast_shapes((4, 2, 3), (2, 1), (1, 3))
+    (4, 2, 3)
+    """
+    if (
+        xp is not None
+        and all(isinstance(size, int) for shape in shapes for size in shape)
```
lucascolley, line 132: for which namespaces is this condition required?

Cyril-36: It is needed for the native paths that do provide `broadcast_shapes`: NumPy, JAX, and PyTorch reject `None` / `math.nan` shape sizes, and CuPy should follow NumPy here. The guard keeps those unknown-size cases on the generic fallback, which preserves the existing behavior of normalizing unknown sizes to `None`. Dask is no longer in this delegation path since it does not provide `broadcast_shapes`.

tests/test_funcs.py
```diff
@@ -489,6 +489,41 @@ def test_5D_values(self, xp: ModuleType):
...
+
+    def test_fallback_for_unknown_sizes(self, monkeypatch: pytest.MonkeyPatch):
+        def mock_broadcast_shapes(*_shapes: tuple[int, ...]) -> tuple[int, ...]:
+            msg = "Native delegation should not handle unknown sizes"
+            raise AssertionError(msg)
+
+        monkeypatch.setattr(np, "broadcast_shapes", mock_broadcast_shapes)
+
+        assert broadcast_shapes((None,), (1,), xp=np) == (None,)
+        assert broadcast_shapes((math.nan,), (1,), xp=np) == (None,)
```
lucascolley: I think we can just remove this test

Cyril-36:
Thanks, removed the redundant unknown-size fallback test.

lucascolley (approved):
thanks @Cyril-36 !

lucascolley:
@all-contributors please add @Cyril-36 for code

PR #720, opened 2026-05-09
ENH: add `check_namespace` to testing utilities

### LINKED ISSUE #719: ENH: Add `check_namespace` option in testing utilities.
Add the option to disable or enable namespace check for `xp_test` in the private test functions. They already have toggles for `dtype`, `shape`, and `scaler`.

prady0t:
Closes #719 
Towards #17


lucascolley (commented):
@prady0t actually, I just went back to gh-267 and read:

> I decided not to add a `check_namespace` toggle: if `check_namespace` is False and the namespaces don't match, other operations might fail. I didn't want to work on fixing this if it's not an important feature. There is only one file in SciPy that uses `check_namespace=False`, and it probably shouldn't.

Maybe we just want to close this?

prady0t:
Ah. Good point indeed. Let's close this.

PR #718, opened 2026-05-07
ENH: add `angle`

prady0t:
Fixes #714

PR #725, opened 2026-05-11
Add the angle function

### LINKED ISSUE #714: ENH: add `angle`
As per Numpy:
```python
def angle(z, deg=False):
    """
    Return the angle of the complex argument.

    Parameters
    ----------
    z : array_like
        A complex number or sequence of complex numbers.
    deg : bool, optional
        Return angle in degrees if True, radians if False (default).

    Returns
    -------
    angle : ndarray or scalar
        The counterclockwise angle from the positive real axis on the complex
        plane in the range ``(-pi, pi]``, with dtype as numpy.float64.

        .. versionchanged:: 1.16.0
            This function works on subclasses of ndarray like `ma.array`.

    See Also
    --------
    arctan2
    absolute

    Notes
    -----
    This function passes the imaginary and real parts of the argument to
    `arctan2` to compute the result; consequently, it follows the convention
    of `arctan2` when the magnitude of the argument is zero. See example.

    Examples
    --------
    >>> import numpy as np
    >>> np.angle([1.0, 1.0j, 1+1j])               # in radians
    array([ 0.        ,  1.57079633,  0.78539816]) # may vary
    >>> np.angle(1+1j, deg=True)                  # in degrees
    45.0
    >>> np.angle([0., -0., complex(0., -0.), complex(-0., -0.)])  # convention
    array([ 0.        ,  3.14159265, -0.        , -3.14159265])

    """
    z = asanyarray(z)
    if issubclass(z.dtype.type, _nx.complexfloating):
        zimag = z.imag
        zreal = z.real
    else:
        zimag = 0
        zreal = z

    a = arctan2(zimag, zreal)
    if deg:
        a *= 180/pi
    return a
```

NeilGirdhar:
Fixes #714

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -6,6 +6,7 @@
     :nosignatures:
     :toctree: generated
 
+    angle
     apply_where
     argpartition
     at
--- a/src/array_api_extra/__init__.py
+++ b/src/array_api_extra/__init__.py
@@ -1,6 +1,7 @@
 """Extra array functions built on top of the array API standard."""
 
 from ._delegation import (
+    angle,
     argpartition,
     atleast_nd,
     cov,
@@ -32,6 +33,7 @@ __version__ = "0.10.2.dev0"
 # pylint: disable=duplicate-code
 __all__ = [
     "__version__",
+    "angle",
     "apply_where",
     "argpartition",
     "at",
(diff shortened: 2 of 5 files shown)


lucascolley:
see also gh-718

NeilGirdhar:
My mistake, feel free to close :smile:

NeilGirdhar:
Actually, if you look carefully at the other one, there are some errors (doesn't handle integers, doesn't handle non float64 floats properly, etc.) 

To my eye, this looks closer to ideal?  But feel free to do what you think is best.

lucascolley:
Would you be willing to review that one?

NeilGirdhar (commented):
(whoops)

lucascolley:
great thanks for reviewing, let's continue over there

Issue #724, opened 2026-05-10
ENH: testing: parameter to toggle comparing a scalar with an array?


prady0t:
Currently, the private xp test functions cannot compare a scaler with an array, even though we are ultimately using `np.testing`, we check for the equality of the total number of elements before:

https://github.com/data-apis/array-api-extra/blob/5045afa282b3956607265b1d6545898b23524278/src/array_api_extra/_lib/_testing.py#L84

To align more with numpy testing behaviour, we can use the strict parameter to check for this. If set to true, do the above check, if False check if it can broadcast. See the numpy example [here](https://numpy.org/devdocs/reference/generated/numpy.testing.assert_array_equal.html#:~:text=As%20mentioned%20in%20the%20Notes%20section%2C%20assert_array_equal%20has%20special%20handling%20for%20scalars.%20Here%20the%20test%20checks%20that%20each%20value%20in%20x%20is%203:). 

An even cleaner solution would be to remove the above check entirely and pass the `strict` parameter directly to the np.testing function call. The current numpy behaviour is different from what's suggested by these comments:

https://github.com/data-apis/array-api-extra/blob/5045afa282b3956607265b1d6545898b23524278/src/array_api_extra/_lib/_testing.py#L81

Either it refers to an older version of numpy testing, or I'm missing something.

lucascolley:
> The current numpy behaviour is different from what's suggested by these comments

Could you clarify how it is different?

prady0t:
The current behaviour is to flatten the array, calculate the total number of elements, and raise an AssertionError if the total number of elements does not match. It does not take into account scalers. We do not consider scalers as inputs (even though the np version has a special case for it) and  this also raises errors when doing

https://github.com/data-apis/array-api-extra/blob/ba2ea5efd6ac14b7506ce96968762193338d5c41/src/array_api_extra/_lib/_testing.py#L68

We can convert the `desired` to `xp.asarray`, but the check for the total number of elements above would still trigger. 

[See](https://numpy.org/devdocs/reference/generated/numpy.testing.assert_array_equal.html#numpy-testing-assert-array-equal) numpy docs example. 

Can we not skip this test and let numpy handle it instead?

lucascolley:
can you show in code an assertion which passes/fails for `np.testing` and the other way around for our assertions?

prady0t:
Try this

```py

import numpy as np
import array_api_strict as xp
import array_api_extra as xpx
[...]

PR #729, opened 2026-05-13
ENH: testing: special case 0D-arrays for `check_shape=False`

prady0t:
Discussed in #724 

Fixing `check_shape`'s else condition to take into account 0d arrays. They are special case in numpy. [See](https://numpy.org/devdocs/reference/generated/numpy.testing.assert_array_equal.html#numpy-testing-assert-array-equal)

### DIFF
--- a/src/array_api_extra/_lib/_testing.py
+++ b/src/array_api_extra/_lib/_testing.py
@@ -81,7 +81,7 @@ def _check_ns_shape_dtype(
         # Ignore shape, but check flattened size. This is normally done by
         # np.testing.assert_array_equal etc even when strict=False, but not for
         # non-materializable arrays.
-        # This check excludes 0d arrays as they are special case in numpy.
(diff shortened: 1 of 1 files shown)


src/array_api_extra/_lib/_testing.py
```diff
@@ -77,10 +77,11 @@ def _check_ns_shape_dtype(
     if check_shape:
         msg = f"shapes do not match: {actual_shape} != f{desired_shape}"
         assert actual_shape == desired_shape, msg
-    else:
+    elif desired.ndim > 0:
         # Ignore shape, but check flattened size. This is normally done by
         # np.testing.assert_array_equal etc even when strict=False, but not for
         # non-materializable arrays.
+        # This check excludes 0d arrays as they are special case in numpy.
```
```suggestion
        # This check excludes 0d arrays as they are special-cased in NumPy.
```

lucascolley (approved):

PR #735, opened 2026-05-13
BUG: testing: fix `check_shape=False` with broadcasting

### LINKED ISSUE #724: ENH: testing: parameter to toggle comparing a scalar with an array?

prady0t:
Discussed in #724

### DIFF
--- a/src/array_api_extra/_lib/_testing.py
+++ b/src/array_api_extra/_lib/_testing.py
@@ -37,7 +37,7 @@ def _check_ns_shape_dtype(
     check_dtype: bool,
     check_shape: bool,
     check_scalar: bool,
-) -> ModuleType:  # numpydoc ignore=RT03
+) -> tuple[Array, Array, ModuleType]:  # numpydoc ignore=RT03
     """
     Assert that namespace, shape and dtype of the two arrays match.
 
@@ -55,7 +55,7 @@ def _check_ns_shape_dtype(
 
     Returns
     -------
-    Arrays namespace.
+    Actual array, desired array, and their namespace.
     """
     actual_xp = array_namespace(actual)  # Raises on scalars and lists
     desired_xp = array_namespace(desired)
@@ -63,6 +63,16 @@ def _check_ns_shape_dtype(
     msg = f"namespaces do not match: {actual_xp} != f{desired_xp}"
     assert actual_xp == desired_xp, msg
 
+    if is_numpy_namespace(actual_xp) and check_scalar:
+        # only NumPy distinguishes between scalars and arrays; we do if check_scalar.
+        _msg = (
+            "array-ness does not match:\n Actual: "
+            f"{type(actual)}\n Desired: {type(desired)}"
+        )
+        assert np.isscalar(actual) == np.isscalar(desired), _msg
+
+    actual = desired_xp.asarray(actual)
+    desired = desired_xp.asarray(desired)
     # Dask uses nan instead of None for unknown shapes
     actual_shape = cast(tuple[float, ...], actual.shape)
     desired_shape = cast(tuple[float, ...], desired.shape)
@@ -90,16 +100,8 @@ def _check_ns_shape_dtype(
     if check_dtype:
         msg = f"dtypes do not match: {actual.dtype} != {desired.dtype}"
         assert actual.dtype == desired.dtype, msg
-
-    if is_numpy_namespace(actual_xp) and check_scalar:
-        # only NumPy distinguishes between scalars and arrays; we do if check_scalar.
-        _msg = (
-            "array-ness does not match:\n Actual: "
-            f"{type(actual)}\n Desired: {type(desired)}"
-        )
-        assert np.isscalar(actual) == np.isscalar(desired), _msg
-
-    return desired_xp
+    desired = desired_xp.broadcast_to(desired, actual.shape)
+    return actual, desired, desired_xp
 
 
 def _is_materializable(x: Array) -> bool:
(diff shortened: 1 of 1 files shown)


src/array_api_extra/_lib/_testing.py
```diff
@@ -55,14 +55,24 @@ def _check_ns_shape_dtype(
...
     msg = f"namespaces do not match: {actual_xp} != f{desired_xp}"
     assert actual_xp == desired_xp, msg
 
+    if is_numpy_namespace(actual_xp) and check_scalar:
+        # only NumPy distinguishes between scalars and arrays; we do if check_scalar.
+        _msg = (
+            "array-ness does not match:\n Actual: "
+            f"{type(actual)}\n Desired: {type(desired)}"
+        )
+        assert np.isscalar(actual) == np.isscalar(desired), _msg
```
prady0t, line 72: This has been moved up before converting to `xp.asarrsy` so that we can still check for scalars

src/array_api_extra/_lib/_testing.py
```diff
@@ -55,14 +55,24 @@ def _check_ns_shape_dtype(
...
+
+    actual = desired_xp.asarray(actual)
+    desired = desired_xp.asarray(desired)
```
prady0t: Instead of relying on the user to give an array input, we convert the input to an array here. In this way, we can still use `check_scalar`, which was not usable before.

src/array_api_extra/_lib/_testing.py
```diff
@@ -55,14 +55,24 @@ def _check_ns_shape_dtype(
 
     Returns
     -------
-    Arrays namespace.
+    Actual array, desired array, and their namespace.
     """
     actual_xp = array_namespace(actual)  # Raises on scalars and lists
```
lucascolley: unrelated to your diff, but this would help clarify since the use of 'scalars' below is rather 'NumPy scalars'
```suggestion
    actual_xp = array_namespace(actual)  # Raises on Python scalars and lists
```

src/array_api_extra/_lib/_testing.py
lucascolley: I disagree — I think `check_scalar` behaves as intended on `main`:

```python
In [1]: import numpy as np

In [2]: import array_api_extra._lib._testing as xpt

In [3]: xpt.xp_assert_equal(np.float64(5.0), np.float64(5.0))

In [4]: xpt.xp_assert_equal(np.float64(5.0), np.asarray(5.0))

In [5]: xpt.xp_assert_equal(np.float64(5.0), np.asarray(5.0), check_scalar=True)
File ~/ghq/github.com/data-apis/array-api-extra/src/array_api_extra/_lib/_testing.py:100, in _check_ns_shape_dtype(actual, desired, check_dtype, check_shape, check_scalar)
     94 if is_numpy_namespace(actual_xp) and check_scalar:
     95     # only NumPy distinguishes between scalars and arrays; we do if check_scalar.
     96     _msg = (
     97         "array-ness does not match:\n Actual: "
     98         f"{type(actual)}\n Desired: {type(desired)}"
     99     )
--> 100     assert np.isscalar(actual) == np.isscalar(desired), _msg
    102 return desired_xp

AssertionError: array-ness does not match:
 Actual: <class 'numpy.float64'>
 Desired: <class 'numpy.ndarray'>
```

I don't blame you for getting confused 😅 since NumPy scalars like `np.float64(5.0)` are not a concept which exists in the standard. `check_scalar` is about those, _not_ Python scalars like `5.0`, which are already rejected on line 60.

```suggestion
```

lucascolley (changes requested):

prady0t:
Ah! My bad.

src/array_api_extra/_lib/_testing.py
```diff
@@ -90,16 +98,8 @@ def _check_ns_shape_dtype(
...
-    if is_numpy_namespace(actual_xp) and check_scalar:
-        # only NumPy distinguishes between scalars and arrays; we do if check_scalar.
-        _msg = (
-            "array-ness does not match:\n Actual: "
-            f"{type(actual)}\n Desired: {type(desired)}"
-        )
-        assert np.isscalar(actual) == np.isscalar(desired), _msg
-
-    return desired_xp
+    desired = desired_xp.broadcast_to(desired, actual.shape)
```
```suggestion
    desired = desired_xp.broadcast_to(desired, actual_shape)
```

src/array_api_extra/_lib/_testing.py
lucascolley: Probably we can move to using `compute_chunk_sizes` here to fix the broadcasting: https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L254-L258

lucascolley (approved):
LGTM now if CI is happy, thanks Pradyot!

PR #743, opened 2026-05-16
ENH: testing: support 0-D arrays for `rtol` and `atol`

prady0t:
As discussed here: https://github.com/scipy/scipy/pull/25143#issuecomment-4468111548

### DIFF
--- a/src/array_api_extra/_lib/_testing.py
+++ b/src/array_api_extra/_lib/_testing.py
@@ -288,16 +288,12 @@ def xp_assert_close(
             rtol = 1e-7
 
     if hasattr(atol, "ndim"):
-        atol = as_numpy_array(atol, xp=xp)
-        if atol.ndim > 0:
-            msg = "atol must be a scalar or 0-D array"
-            raise TypeError(msg)
+        if atol.ndim == 0:
+            atol = as_numpy_array(atol, xp=xp)
 
     if hasattr(rtol, "ndim"):
-        rtol = as_numpy_array(rtol, xp=xp)
-        if rtol.ndim > 0:
-            msg = "rtol must be a scalar or 0-D array"
-            raise TypeError(msg)
+        if rtol.ndim == 0:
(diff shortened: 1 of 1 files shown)


src/array_api_extra/_lib/_testing.py
```diff
@@ -246,9 +246,9 @@ def xp_assert_close(
         The array produced by the tested function.
     desired : Array
         The expected array (typically hardcoded).
-    rtol : float, optional
+    rtol : float or Array, optional
         Relative tolerance. Default: dtype-dependent.
-    atol : float, optional
+    atol : float or Array, optional
```
lucascolley, line 251: Let's wait for feedback on https://github.com/numpy/numpy/issues/31449 before we proceed, but it would be good to specify here the supported dimensionality

lucascolley (commented):
Nice, feel free to update your PR to SciPy to see what passes/fails with this change.

lucascolley (commented):
For the CI failures here, I think we should update `as_numpy_array` to be a no-op on NumPy scalars.

lucascolley:
On second thought, seems like we can simplify this by just calling `float` on the input — want to see if this works for SciPy?

src/array_api_extra/_lib/_testing.py
```diff
@@ -286,13 +288,18 @@ def xp_assert_close(
...
+
+    atol = float(atol)
 
     actual_np = as_numpy_array(actual, xp=xp)
     desired_np = as_numpy_array(desired, xp=xp)
     np.testing.assert_allclose(  # pyright: ignore[reportCallIssue]
         actual_np,
         desired_np,
-        rtol=rtol,  # pyright: ignore[reportArgumentType]
+        # https://github.com/numpy/numpy/issues/31449
```
```suggestion
```

prady0t:
> On second thought, seems like we can simplify this by just calling float on the input 

Yeah, that's what I initially suggested. Let's merge this.

PR #747, opened 2026-05-18
ENH: testing: add `xp` param to assertion functions

prady0t:

### DIFF
--- a/src/array_api_extra/_lib/_testing.py
+++ b/src/array_api_extra/_lib/_testing.py
@@ -8,6 +8,9 @@ See also ..testing for public testing utilities.
 from __future__ import annotations
 
 import math
+from collections.abc import Generator
+from contextlib import contextmanager
+from contextvars import ContextVar
 from types import ModuleType
 from typing import Any, cast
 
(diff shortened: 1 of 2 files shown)


lucascolley (commented):
thanks Pradyot! Could you update the SciPy branch to use this branch?

prady0t:
> Could you update the SciPy branch to use this branch?

Will have to merge this first right?

lucascolley:
> > Could you update the SciPy branch to use this branch?
> 
> Will have to merge this first right?

Ah, no! I think the easiest way is to run `gh pr checkout 747` from within the submodule.

prady0t:
We till can't remove these parts from scipy for two reasons:
- We need a fallback to find XP if not given before converting scalar to array (something we do not do here)
- `xp_assert_close_nulp` exists and uses it too.

@lucascolley What do you think?

lucascolley:
> We need a fallback to find XP if not given before converting scalar to array (something we do not do here)

Could you explain this in more detail?

prady0t:
Currently, we do not allow for scalar or Python lists as input, but Scipy uses them everywhere. For this reason, we have to convert them to their respective `xp.asarray`, but to do that, we need to know' xp' in advance. Hence, we use this block before `_convert_scalar_to_array`:

```py
    if xp is None:
        try:
            xp = _default_xp_ctxvar.get()
        except LookupError:
            xp = array_namespace(desired)
```

Hence, we need this block before we convert to xp.asarray and send it to our xp assert functions.

src/array_api_extra/_lib/_testing.py
```diff
@@ -60,8 +103,12 @@ def _check_ns_shape_dtype(
     actual_xp = array_namespace(actual)  # Raises on Python scalars and lists
     desired_xp = array_namespace(desired)
 
-    msg = f"namespaces do not match: {actual_xp} != f{desired_xp}"
-    assert actual_xp == desired_xp, msg
+    if xp is None:
+        try:
+            xp = _default_xp_ctxvar.get()
+        except LookupError:
+            xp = array_namespace(desired)
```
prady0t: We can remove the `default_xp` context manager code from scipy and keep it here (and import from here), but we would still do this before converting the inputs to asarrays in scipy. On another thought, we can wrap this around a function and expose it.

lucascolley:
> we have to convert them to their respective `xp.asarray`, but to do that, we need to know' xp' in advance. Hence, we use this block before `_convert_scalar_to_array`

yeah, this makes me think that maybe the context manager should stay on the SciPy side for now. Let me see if I can get that working.

lucascolley:
If the SciPy PR is green with this then I think we merge

PR #749, opened 2026-05-19
ENH: testing: add `dlpack` conversion support

### LINKED ISSUE #748: Should we use DLPack for numpy conversions of arrays that support it?
Inspired by: https://github.com/wilsonrljr/sysidentpy/blob/5877e8e452bda46f8007421effb5bea1bff2fa91/sysidentpy/_lib/_array_api.py#L199

We can use DLPack minimally if the array object supports it.

prady0t:
Closes: #748


lucascolley (approved):
I'm happy with this once the linter is happy

PR #775, opened 2026-06-03
CD: use `ubuntu-latest` for publishing dist

lucascolley:


.github/workflows/cd.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       id-token: write # for trusted publishing
       attestations: write # for trusted publishing
       contents: read
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-latest
```
paddyroddy, line 40: I've fallen for this before :(

PR #785, opened 2026-06-09
API: testing: relax condition that `desired` match `xp` in assertions

lucascolley:
this is technically breaking so may as well milestone this for `0.11.0`


prady0t:
~~Also, I believe this deviates from scipy's behaviour of inferring namespace from desired.~~

My bad, didn't see the suggested diff in SciPy

Issue #789, opened 2026-06-09
ENH: add support for `nanmin`, `nanmax`, `nanmean` and `nansum`


OmarManzoor:
Would it make sense to include these

- [x] **nanmin**
- [x] **nanmax**
- [x] **nanmean**
- [x] **nansum**

There is a corresponding issue opened in https://github.com/data-apis/array-api/issues/621 but I don't think that has been planned yet.

scikit-learn reference: [array api nan reduction utils](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_array_api.py#L925-L990)

What do you think @lucascolley

lucascolley:
feel free to add to the agenda at https://hackmd.io/zn5bvdZTQIeJmb3RW1B-8g (even if you can't make it to the meeting it could still be discussed)

OmarManzoor:
@lucascolley Would it be okay if you could add it please? Because otherwise editing the agenda that you shared requires signing up and I don't really use this platform.

lucascolley:
This was discussed at https://hackmd.io/zn5bvdZTQIeJmb3RW1B-8g#nan-reductions-Lucas — a PR adding these functions to array-api-extra would be welcome!

PR #804, opened 2026-06-16
ENH: add `nanmin`

OmarManzoor:
Towards: #789

- Adds array API support for `nanmin`


OmarManzoor:
@lucascolley I added one of these for a start. I think after we finalize one of them the others can follow along.

src/array_api_extra/_lib/_funcs.py
```diff
@@ -757,3 +757,24 @@ def angle(z: Array, /, *, deg: bool = False, xp: ModuleType | None = None) -> Ar
     if deg:
         a = a * 180 / xp.pi
     return a
+
+
+def nanmin(  # numpydoc ignore=PR01,RT01
```
lucascolley, line 829: whereabouts did you take this implementation from?

lucascolley (commented):
there are a few merge conflicts

src/array_api_extra/_lib/_funcs.py
OmarManzoor: `scikit-learn` directly. As for how that is implemented there I am not exactly sure.

lucascolley: would be good to check the history there

OmarManzoor: https://github.com/scikit-learn/scikit-learn/pull/26243/changes

This is the actual PR

lucascolley: thanks — @betatim would you mind weighing in on how this implementation arose?

EDIT: /feel free to review also :)

OmarManzoor: I think this mainly follows the numpy implementation ignoring cases of objects or ndarray specifics:
https://github.com/numpy/numpy/blob/main/numpy/lib/_nanfunctions_impl.py#L252-L374

betatim: What exactly are you looking for in terms of histroy Lucas?

My guess is that like Omar said it is based on what is in Numpy, potentially minus some extra args/handling of edge cases.

lucascolley: > What exactly are you looking for in terms of histroy Lucas?
> 
> 
> 
> My guess is that like Omar said it is based on what is in Numpy, potentially minus some extra args/handling of edge cases.

If that's the case, then yeah, I'll start by looking at what NumPy does

betatim: Yeah, the way these functions make their way into scikit-learn is "this exists in numpy, doesn't exist in array API, we need it, let's translate the numpy code"

src/array_api_extra/_delegation.py
```diff
@@ -1378,3 +1379,44 @@ def unravel_index(
...
+    ----------
+    a : Array
+        Input array.
+
+    axis : int or tuple of ints or None, optional
+        Axis or axes along which the minimum is computed. The default is to compute
+        the minimum of the flattened array.
+
+    xp : array_namespace, optional
+        The standard-compatible namespace for `x`. Default: infer.
```
lucascolley: note change of variable name from `x` to `a`
```suggestion
    a : Array
        Input array.
    axis : int or tuple of ints or None, optional
        Axis or axes along which the minimum is computed. The default is to compute
        the minimum of the flattened array.
    xp : array_namespace, optional
        The standard-compatible namespace for `a`. Default: infer.
```

src/array_api_extra/_delegation.py
```diff
@@ -1378,3 +1379,44 @@ def unravel_index(
...
+
+    Returns
+    -------
+    array
+        An array of minimum values along the given axis, ignoring NaNs.
+    """
```
lucascolley, line 1625: can we add an example?

tests/test_funcs.py
```diff
@@ -2044,3 +2045,71 @@ def test_xp(self, xp: ModuleType):
         res = unravel_index(indices, shape, xp=xp)
         for res_arr, exp_arr in zip(res, expected, strict=True):
             assert_equal(res_arr, exp_arr)
+
+
+class TestNanMin:
```
lucascolley, line 2209: any other tests we can grab from https://github.com/numpy/numpy/blob/main/numpy/lib/tests/test_nanfunctions.py ?

tests/test_funcs.py
```diff
@@ -2044,3 +2045,71 @@ def test_xp(self, xp: ModuleType):
...
+        res = nanmin(a)
+        expected = 0.1
+        assert res == expected
+
+        a = xp.asarray([0.1, 1.0, xp.nan, -xp.inf])
+        res = nanmin(a)
+        expected = -xp.inf
+        assert res == expected
+
+    @pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
lucascolley: can we make this slightly more specific, e.g. to include `nan` in the string

OmarManzoor: You mean the warning message?

lucascolley: Yeah

tests/test_funcs.py
OmarManzoor: Added two more tests:

- test with a simple scalar
- test with using the xp param

OmarManzoor:
@lucascolley Does this look fine to merge?

tests/test_funcs.py
```diff
@@ -2203,3 +2204,84 @@ def test_xp(self, xp: ModuleType):
...
+            ]
+        )
+
+        res = nanmin(a, axis=0, xp=xp)
+        expected = xp.asarray([xp.nan, 2.0])
+        assert_equal(res, expected)
+
+    @pytest.mark.skip_xp_backend(
+        Backend.TORCH, reason="Tensor.item() cannot be called on meta tensors"
+    )
```
lucascolley, line 2273: can we change this reason to point to the part of torch we are _using_ that is lacking support? E.g. (if this is accurate) "`torch.nanmin` does not support tensors on meta device".

OmarManzoor: Done! I think this seems better.

lucascolley (approved):
thanks Omar! nice work

lucascolley:
I think we should test this out in scikit-learn before adding more of these nan functions here, btw

OmarManzoor:
> I think we should test this out in scikit-learn before adding more of these nan functions here, btw

Maybe we can add `nanmax` as well and then test these two out together? What do you think? I think they are quite similar.

lucascolley:
> > I think we should test this out in scikit-learn before adding more of these nan functions here, btw
> 
> Maybe we can add `nanmax` as well and then test these two out together? What do you think? I think they are quite similar.

feel free to start working on a PR!

PR #832, opened 2026-07-05
ENH: pad: normalize `pad_width` on the host for torch

rgommers:
`pad_width` is host metadata (an int or (before, after) pairs), but the torch delegation branch normalized it by round-tripping through a tensor: `xp.asarray(pad_width)` places it on the *default* device, and the `tuple(...)` call to build the python ints that `torch.nn.functional.pad` expects then performs one device-to-host transfer per element. With a CUDA default device that is a needless host-to-device copy plus 2*ndim synchronizing `.item()` reads per call; with a data-free `meta` default device it raises outright.

Factor the pure-python normalization out of `_funcs.pad` into `normalize_pad_width` and reuse it in the torch branch, reversing the axis order on the host.

This also aligns the torch branch with the documented `pad_width` types (the tensor broadcast incidentally accepted undocumented forms such as shape ``(n, 1)``).

Found this when working on non-default device handling issues in `scipy`; SciPy tests will cover this so I won't worry about introducing multi-device testing here right now.

AI Generation Disclosure: patch generated with Claude Fable 5 (had to tweak both code comments and commit message a bit, but the hard work was done by Claude here).

### DIFF
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -796,7 +796,11 @@ def pad(
         # `torch/_numpy`'s implementation (avoids device transfers)
         pad_width_seq = _funcs.normalize_pad_width(pad_width, x.ndim)
         # torch.nn.functional.pad counts dimensions from the last one
-        flat_pad_width = [w for pair in reversed(pad_width_seq) for w in pair]
+        flat_pad_width = [
+            w
+            for pair in reversed(pad_width_seq)
+            for w in pair
+        ]
         return xp.nn.functional.pad(x, tuple(flat_pad_width), value=constant_values)
 
     return _funcs.pad(x, pad_width, constant_values=constant_values, xp=xp)
--- a/src/array_api_extra/_delegation.py
+++ b/src/array_api_extra/_delegation.py
@@ -791,12 +791,13 @@ def pad(
     ):
         return xp.pad(x, pad_width, mode, constant_values=constant_values)
 
-    # https://github.com/pytorch/pytorch/blob/cf76c05b4dc629ac989d1fb8e789d4fac04a095a/torch/_numpy/_funcs_impl.py#L2045-L2056
     if is_torch_namespace(xp):
-        pad_width = xp.asarray(pad_width)
-        pad_width = xp.broadcast_to(pad_width, (x.ndim, 2))
-        pad_width = xp.flip(pad_width, axis=(0,)).flatten()
-        return xp.nn.functional.pad(x, tuple(pad_width), value=constant_values)  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]
+        # normalize `pad_width` on the host rather than through a tensor as done in
+        # `torch/_numpy`'s implementation (avoids device transfers)
+        pad_width_seq = _funcs.normalize_pad_width(pad_width, x.ndim)
+        # torch.nn.functional.pad counts dimensions from the last one
+        flat_pad_width = [w for pair in reversed(pad_width_seq) for w in pair]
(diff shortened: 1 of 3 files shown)


src/array_api_extra/_lib/_funcs.py
```diff
@@ -538,6 +538,22 @@ def nunique(x: Array, /, *, xp: ModuleType | None = None) -> Array:
     )
 
 
+def normalize_pad_width(
```
lucascolley: Arguably this belongs in `_helpers` instead

src/array_api_extra/_delegation.py
```diff
@@ -792,12 +791,13 @@ def pad(
...
     if is_torch_namespace(xp):
-        pad_width = xp.asarray(pad_width)
-        pad_width = xp.broadcast_to(pad_width, (x.ndim, 2))
-        pad_width = xp.flip(pad_width, axis=(0,)).flatten()
-        return xp.nn.functional.pad(x, tuple(pad_width), value=constant_values)  # type: ignore[arg-type]  # pyright: ignore[reportArgumentType]
+        # normalize `pad_width` on the host rather than through a tensor as done in
+        # `torch/_numpy`'s implementation (avoids device transfers)
+        pad_width_seq = _funcs.normalize_pad_width(pad_width, x.ndim)
+        # torch.nn.functional.pad counts dimensions from the last one
+        flat_pad_width = [w for pair in reversed(pad_width_seq) for w in pair]
```
lucascolley: non-blocking but something like this makes this expression easier to parse to my eyes
```suggestion
        flat_pad_width = [
            w
            for pair in reversed(pad_width_seq)
            for w in pair
        ]
```

lucascolley:
> SciPy tests will cover this so I won't worry about introducing multi-device testing here right now.

Would be good to at least open an issue for this

rgommers:
It's already covered by the generic issue for non-default devices; it's one of very many (see gh-25567) so not worth a separate issue.

src/array_api_extra/_delegation.py
rgommers: sure, let's apply that change.

lucascolley:
> It's already covered by the generic issue for non-default devices; it's one of very many (see gh-25567) so not worth a separate issue. 

Sorry, I meant an issue on this repo

src/array_api_extra/_lib/_funcs.py
rgommers: I'll move it if you prefer that - but are you sure? It's specific to just `pad` and not useful as a standalone generic helper, so keeping it right above `def pad:` would be my choice.

rgommers:
Ah, that seems reasonable - a generic "introduce multi-device or non-default device testing for array-api-extra" issue. I think after we have it for SciPy, it might be clearer if/how it can be done in this repo.

src/array_api_extra/_lib/_funcs.py
lucascolley: Given that it is used in 2 separate modules, I would rather move it to `_helpers`, yes. Let's add it to `__all__` over there, too.

lucascolley (approved):

rgommers:
Hmm, looks like `torch` isn't tested in CI, since only the linter failed.

lucascolley:
> Hmm, looks like `torch` isn't tested in CI, since only the linter failed.

it is tested otherwise coverage would fail... I think the previous commit still worked due to the helper being imported into `_funcs`. That is what the error code suggests at least: `reportPrivateLocalImportUsage`

lucascolley:
thanks Ralf!

Issue #833, opened 2026-07-06
TST: non-default device testing


lucascolley:
Once SciPy has figured out a decent strategy for this (https://github.com/scipy/scipy/pull/25565), we should do the same, and potentially also extract anything that fits into `xpx.testing`

prady0t:
One thing that we can add right now is the `check_device` parameter here:

https://github.com/data-apis/array-api-extra/blob/eb8ff5dc6ff906e97d6e065caf87cf31898f898a/src/array_api_extra/testing.py#L567

Inspired by: https://github.com/scipy/scipy/pull/25565

PR #835, opened 2026-07-06
ENH: testing: Add `check_device` parameter

prady0t:
Towards: #833

### DIFF
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -1055,13 +1055,13 @@ class TestIsClose:
         b = 1
         res = isclose(a, b)
         assert get_device(res) == device
-        assert_equal(res, xp.asarray([False, False, False, False, True]))
+        assert_equal(res, xp.asarray([False, False, False, False, True], device=device))
 
         a = 0.1
         b = xp.asarray([0.01, 0.5, 0.8, 0.9, 0.100001], device=device, dtype=xp.float64)
         res = isclose(a, b)
         assert get_device(res) == device
-        assert_equal(res, xp.asarray([False, False, False, False, True]))
+        assert_equal(res, xp.asarray([False, False, False, False, True], device=device))
 
 
 class TestKron:
@@ -1676,7 +1676,7 @@ class TestIsIn:
 
         a = xp.asarray([0, 3, 6, 10], device=device)
         b = xp.asarray([1, 2, 3, 10], device=device)
-        expected = xp.asarray([True, False, True, False])
+        expected = xp.asarray([True, False, True, False], device=device)
         res = isin(a, b, assume_unique=True, invert=True)
         assert get_device(res) == device
         assert_equal(res, expected)
--- a/tests/test_funcs.py
+++ b/tests/test_funcs.py
@@ -211,15 +211,6 @@ class TestApplyWhere:
         expect = xp.where(cond, self.f1(x), self.f2(x))
         assert_equal(actual, expect)
 
-    def test_device(self, xp: ModuleType, device: Device):
-        x = xp.asarray([1, 2, 3, 4], device=device)
-        y = apply_where(x % 2 == 0, x, self.f1, self.f2)
-        assert get_device(y) == device
-        y = apply_where(x % 2 == 0, x, self.f1, fill_value=0)
-        assert get_device(y) == device
-        y = apply_where(x % 2 == 0, x, self.f1, fill_value=x)
-        assert get_device(y) == device
-
     @pytest.mark.filterwarnings("ignore::RuntimeWarning")  # overflows, etc.
     @hypothesis.settings(
         # The xp and library fixtures are not regenerated between hypothesis iterations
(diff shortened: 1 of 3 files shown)


tests/test_funcs.py
```diff
@@ -1054,14 +1054,12 @@ def test_array_on_device_with_scalar(self, xp: ModuleType, device: Device):
         a = xp.asarray([0.01, 0.5, 0.8, 0.9, 1.00001], device=device, dtype=xp.float64)
         b = 1
         res = isclose(a, b)
-        assert get_device(res) == device
-        assert_equal(res, xp.asarray([False, False, False, False, True]))
+        assert_equal(res, xp.asarray([False, False, False, False, True], device=device))
```
prady0t, line 1058: ~This was the only conflicting test.~

tests/test_funcs.py
```diff
@@ -1676,7 +1674,7 @@ def test_assume_unique_and_invert(
 
         a = xp.asarray([0, 3, 6, 10], device=device)
         b = xp.asarray([1, 2, 3, 10], device=device)
-        expected = xp.asarray([True, False, True, False])
+        expected = xp.asarray([True, False, True, False], device=device)
```
prady0t, line 1679: Need to specify device here; otherwise, it fails in the assert_equal statement below with: 

```
Actual: array_api_strict.Device('device1')
Desired: array_api_strict.Device('CPU_DEVICE')
```
 The reason is that `res.device` is set by the pytest fixture (as device1) and `expected.device` is set by `array_api_strict` default unless specified.

tests/test_funcs.py
prady0t, line 1058: Same reason as above.

prady0t:
CC @j-bowhay

tests/test_funcs.py
```diff
@@ -1676,7 +1674,7 @@ def test_assume_unique_and_invert(
...
         res = isin(a, b, assume_unique=True, invert=True)
         assert get_device(res) == device
```
j-bowhay, line 1681: With the changes in this PR these sorts of checks can be removed right?

j-bowhay (commented):
It would be nice to have some tests :)

prady0t:
Some irrelevant device checks have been removed as assert functions check for the device by default now. I've also added checks for non-default device.

tests/test_funcs.py
prady0t: Makes sense to remove here and a few more places, but it would still be required to have  `test_device` tests for every function, as it checks whether the function is preserving the original device or not. Example:

```py
    @pytest.mark.xfail_xp_backend(Backend.TORCH, reason="array-api-extra#455")
    def test_device(self, xp: ModuleType, device: Device):
        assert get_device(cov(x)) == device
```

tests/test_funcs.py
```diff
@@ -211,15 +211,6 @@ def test_xp(self, xp: ModuleType):
...
         assert_equal(actual, expect)
 
-    def test_device(self, xp: ModuleType, device: Device):
-        x = xp.asarray([1, 2, 3, 4], device=device)
-        y = apply_where(x % 2 == 0, x, self.f1, self.f2)
-        assert get_device(y) == device
-        y = apply_where(x % 2 == 0, x, self.f1, fill_value=0)
-        assert get_device(y) == device
-        y = apply_where(x % 2 == 0, x, self.f1, fill_value=x)
-        assert get_device(y) == device
```
lucascolley, line 221: hmm, I'm not convinced that these tests are redundant yet. The `device` fixture gives us a non-default device where possible: https://github.com/data-apis/array-api-extra/blob/064cbf742cc6d296f8b2837db04f049f9e4cfd95/tests/conftest.py#L222-L237

IIUC dropping the tests removes that testing coverage of non-default devices. I don't think we should do that until we have an improved solution.

lucascolley (changes requested):

tests/test_funcs.py
lucascolley: gentle bump @prady0t :)

prady0t: You are right. 

With the additional parameter in the assert statements, what we can do is remove the dedicated `test_device` from test classes and just use the `Device` fixture in something like a `test_basic` (or any other existing test), and pass `check_device=True` in the assert test function. This might help clean things up a bit, but it can also introduce some ambiguity if the device test fails.

Apart from that, the test suite does a good job indeed in testing with non-default devices.

I still believe that the `check_device` parameter is a good addition to the `array_api_extra`'s assert functions, as downstream projects can leverage it.

lucascolley: > I still believe that the `check_device` parameter is a good addition to the `array_api_extra`'s assert functions, as downstream projects can leverage it.

yes, for sure!

Maybe it is best to keep the `test_device` methods, though? Line 2 of https://peps.python.org/pep-0020/#the-zen-of-python ...

prady0t: >  Line 2 of https://peps.python.org/pep-0020/#the-zen-of-python ...

Always a charm to read. 

Let's keep them.

tests/test_testing.py
```diff
@@ -206,13 +209,25 @@ def test_device(self, xp: ModuleType, device: Device, func: Callable[..., None])
...
+        default_device = get_device(xp.empty(0))
+        if device != default_device:
+            d = xp.asarray([2], device=default_device)
+            with pytest.raises(AssertionError, match="Devices do not match"):
+                func(a, d)
+            func(a, d, check_device=False)
+            if getattr(device, "type", None) != "meta":
+                e = xp.asarray([0], device=default_device)
+                with pytest.raises(AssertionError, match=self.np_err_msg(func)):
+                    func(a, e, check_device=False)
```
prady0t: This is probably rubbish. The only question I have is: how was this function testing device if assert functions didn't account for it earlier? With the new parameter, it has been set as the default behaviour.

lucascolley: I believe it was testing that alternative devices are *accepted* and don't break the code, although it wasn't testing that the return device was as expected

lucascolley (approved):
great, thanks both!

PR #847, opened 2026-07-13
ENH: add `nanmax`

OmarManzoor:
Towards #789

- Adds support for `nanmax`


lucascolley:
as mentioned earlier, would be nice to see a branch with this and `nanmin` working in sklearn

OmarManzoor:
How should we test? Will you merge this and then I need to adjust the version of `array-api-extra` in `scikit-learn` on my local system?

lucascolley:
you can checkout this branch of array-api-extra inside sklearn

OmarManzoor:
The CI in `scikit-learn` is 🟢

OmarManzoor:
@lucascolley Can we merge this one?

tests/test_funcs.py
```diff
@@ -2285,3 +2286,84 @@ def test_xp(self, axis: int | None, expected_list: list[float], xp: ModuleType):
         res = nanmin(a, axis=axis, xp=xp)
         expected = xp.asarray(expected_list)
         assert_equal(res, expected)
+
+
+class TestNanMax:
```
lucascolley, line 2291: can we reduce the duplication with `TestNanMin` a bit?

lucascolley (changes requested):

tests/test_funcs.py
OmarManzoor: What do you propose?

lucascolley: hmm, I figured we could define test cases at the class level and parametrize the class with both functions, but maybe that is more effort than it is worth given the different data for each.

OmarManzoor: We could try to keep the input data the same but the answers will need to be parametrized as they will be different for both functions

lucascolley: If you want to try something then go for it, otherwise I won't block, I guess this is clean enough already :)

OmarManzoor: If I add parametrizations for each of tests it might get a bit awkward adding functions + axis + expected results in the parameters. In my quick overview I am not sure how that will look as compared to the current separation which doesn't have any long parametrizations.

lucascolley: okay yeah let's leave it as is

lucascolley (approved):
thanks Omar!

PR #890, opened 2026-08-11
deps: exclude Pixi deps newer than 7d old

lucascolley:


lucascolley:
The test failures are caused by upgrading `jaxlib` from `0.10.1` to `0.10.2`: https://github.com/jax-ml/jax/issues/39887

PR #910, opened 2026-08-12
ENH: add `nanmean`

ushnah:
Towards #789 
- add support for nanmean

PR #920, opened 2026-08-13
DOC: organise API reference

### LINKED ISSUE #322: DOC: organise API reference
We have a bit of a mish-mash of functions in the main API reference page at the moment. This is fine for now, but as the API grows it would be helpful to break them up into meaningful categories, e.g. 'creation functions', like the standard does.

Johnkothapalli:
## Summary

Organises the flat assorted API autosummary into categories aligned with the Python Array API standard, while keeping a single page and preserving the generated reference URLs. All 31 existing public entries remain listed exactly once.

Closes #322.

The open #910 adds `nanmean` to this file. If it lands first, I will rebase and place `nanmean` under Statistical functions.

## Validation

- `pixi run docs` (Sphinx with `-E -W`)
- `pixi run lint`
- coverage check confirming 31 API entries, each listed exactly once

AI assistance: Codex helped analyze the grouping and prepare the change; the resulting diff was verified with the project checks above.

### DIFF
--- a/docs/api-assorted.md
+++ b/docs/api-assorted.md
@@ -1,5 +1,30 @@
 # Assorted functions
 
+## Creation functions
+
+```{eval-rst}
+.. currentmodule:: array_api_extra
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    create_diagonal
+    one_hot
+```
+
+## Inspection
+
+```{eval-rst}
+.. currentmodule:: array_api_extra
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    default_dtype
+```
+
+## Element-wise functions
+
 ```{eval-rst}
 .. currentmodule:: array_api_extra
 .. autosummary::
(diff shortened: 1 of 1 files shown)


docs/api-assorted.md
```diff
@@ -1,40 +1,132 @@
...
+## Element-wise functions
+
+```{eval-rst}
+.. currentmodule:: array_api_extra
+.. autosummary::
+    :nosignatures:
+    :toctree: generated
+
+    angle
+    apply_where
```
lucascolley, line 35: `apply_where` is the only one where the category doesn't feel like a great fit to me, but I think it is still arguably the best place for it, so happy to leave it here.

docs/api-assorted.md
```diff
@@ -1,40 +1,132 @@
 # Assorted functions
 
+## Creation functions
```
lucascolley, line 3: now that we have this, I think it would be nice to have a multi-level TOC which splits each category into a separate sub-page. I can take a look at doing that in a follow-up.

lucascolley (approved):
Thanks @Johnkothapalli, this seems smoother than I anticipated!

Another opportunity this presents for a follow-up is to reorganise the implementation files accordingly. I'll take a look at that soon.

PR #963, opened 2026-09-01
ENH: add `mparray` backend

mdhaber:
#### Reference issue
Toward scipy/scipy#24840

#### What does this implement/fix?
This adds MPArray as a tested backend of array-api-extra.

#### Additional information
Skipping CI until conda-forge has the latest MPArray release (v0.2.0 recently released on PyPI), but tests are passing locally with the editable install.

Are there other things I should do (e.g. update documentation somewhere)?

#### AI Generation Disclosure
I told ChatGPT to work mparray into the testing. It made some changes so I could see what was needed, then I changed almost everything in some way.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -373,7 +373,7 @@ numpy = "=1.24.1"
 pytorch = ">=2.12.0"
 dask-core = ">=2026.7.1"  # No distributed, tornado, etc.
 sparse = ">=0.19.2"
-mparray = ">=0.1.0"
+mparray = ">=0.2.2"
 
 [feature.backends.target.unix.dependencies]
 jax = ">=0.10.2"  # waiting for conda-forge/jaxlib-feedstock#326
--- a/pixi.toml
+++ b/pixi.toml
@@ -17,6 +17,7 @@ exclude-newer = "7d"
 
 [exclude-newer]
 pixi-build-python = "5d"  # for compatibility with pixi-build-api-version 7
+mparray = "0d"
 
 ### array-api-extra package definition ###
 
@@ -372,6 +373,7 @@ numpy = "=1.24.1"
 pytorch = ">=2.12.0"
 dask-core = ">=2026.7.1"  # No distributed, tornado, etc.
 sparse = ">=0.19.2"
+mparray = ">=0.1.0"
 
 [feature.backends.target.unix.dependencies]
 jax = ">=0.10.2"  # waiting for conda-forge/jaxlib-feedstock#326
--- a/src/array_api_extra/_set.py
+++ b/src/array_api_extra/_set.py
@@ -101,7 +101,7 @@ def nunique(x: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
     if (
         _compat.is_numpy_namespace(xp)
         or _compat.is_cupy_namespace(xp)
-        or xp.__name__ == "mparray"
+        or _helpers.is_mparray_namespace(xp)
         or (
             _compat.is_torch_namespace(xp)
             and _helpers.capabilities(xp, x)["data-dependent shapes"]
--- a/src/array_api_extra/_set.py
+++ b/src/array_api_extra/_set.py
@@ -101,6 +101,7 @@ def nunique(x: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
     if (
         _compat.is_numpy_namespace(xp)
         or _compat.is_cupy_namespace(xp)
+        or xp.__name__ == "mparray"
         or (
             _compat.is_torch_namespace(xp)
             and _helpers.capabilities(xp, x)["data-dependent shapes"]
--- a/tests/main/test_at.py
+++ b/tests/main/test_at.py
@@ -246,7 +246,7 @@ def test_incompatible_dtype(
             with pytest.warns(FutureWarning, match="cannot safely cast"):
                 z = at_op(x, idx, op, 1.1, copy=copy)
 
-    elif library.like(Backend.DASK):
+    elif library.like(Backend.DASK) or (library == Backend.MPARRAY):
         z = at_op(x, idx, op, 1.1, copy=copy)
 
     elif library.like(Backend.ARRAY_API_STRICT):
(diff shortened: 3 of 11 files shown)


tests/main/test_elementwise.py
```diff
@@ -662,6 +662,7 @@ def test_basic(self, xp: ArrayNamespace):
             atol=1e-11,
         )
 
+    @pytest.mark.skip_xp_backend(Backend.MPARRAY, reason="negative zero not supported")
```
mdhaber, line 665: This is just a fundamental limitation of `mpmath`, it seems. I don't think fixing it is high priority.

tests/main/test_lazy.py
```diff
@@ -27,6 +27,7 @@
                     Backend.TORCH_GPU, reason="device->host copy"
                 ),
                 pytest.mark.skip_xp_backend(Backend.SPARSE, reason="densification"),
+                pytest.mark.skip_xp_backend(Backend.MPARRAY, reason="precision loss"),
```
mdhaber, line 30: I'm not sure how `as_numpy=True` should work. I think `np.asarray(my_mparray)` currently wraps `my_mparray` as the only element of an object array. Instead, it could:
- Extract the underlying object array, but I don't think that's what this is supposed to do.
- Converted the underlying object array to a numerical NumPy array, but it risks precision loss.
- Raise.

tests/main/test_testing.py
```diff
@@ -362,6 +362,7 @@ def test_lazy_xp_function_deprecated_static_argnames():
 
 
 @pytest.mark.skip_xp_backend(Backend.TORCH_GPU, reason="device->host copy")
+@pytest.mark.skip_xp_backend(Backend.MPARRAY, reason="mparray lacks __array_ufunc__")
```
mdhaber, line 365: I don't think we want compiled ufuncs to work on these if they don't preserve the precision; it's probably better to fail. (However, `mparray.special` will provide a lot of special functions, and `scipy.special` will be able to delegate to them.)

pixi.toml
```diff
@@ -17,6 +17,7 @@ exclude-newer = "7d"
 
 [exclude-newer]
 pixi-build-python = "5d"  # for compatibility with pixi-build-api-version 7
+mparray = "0d"
```
mdhaber, line 20: Since I just released this.

pixi.toml
```diff
@@ -372,6 +373,7 @@ numpy = "=1.24.1"
 pytorch = ">=2.12.0"
 dask-core = ">=2026.7.1"  # No distributed, tornado, etc.
 sparse = ">=0.19.2"
+mparray = ">=0.1.0"
```
mdhaber: 0.1.0 is OK, but there would be a few test failures. Maybe should update to 0.2.0.

tests/main/test_at.py
```diff
@@ -246,7 +246,7 @@ def test_incompatible_dtype(
             with pytest.warns(FutureWarning, match="cannot safely cast"):
                 z = at_op(x, idx, op, 1.1, copy=copy)
 
-    elif library.like(Backend.DASK):
+    elif library.like(Backend.DASK) or (library == Backend.MPARRAY):
         z = at_op(x, idx, op, 1.1, copy=copy)
```
mdhaber, line 251: MPArray is currently a little too flexible about setting with higher-precision objects, e.g.
```python3
import mparray as xp
x = xp.asarray([0, 1, 2, 3])
x[0] = xp.asarray(0, dtype=xp.complex64)
x
# MPArray([mpc(real='0.0', imag='0.0'), 1, 2, 3], dtype=int64)
```

See mdhaber/mparray#21.

src/array_api_extra/_set.py
```diff
@@ -101,6 +101,7 @@ def nunique(x: Array, /, *, xp: ArrayNamespace | None = None) -> Array:
     if (
         _compat.is_numpy_namespace(xp)
         or _compat.is_cupy_namespace(xp)
+        or xp.__name__ == "mparray"
```
lucascolley: please could we add a helper for this in `_lib._helpers`

tests/main/test_at.py
lucascolley: Can we add a comment if this branch is potentially temporary?

lucascolley (commented):
great to see how small a diff this requires :)

lucascolley:
why does `mparray` require Python >=3.13? If that has to stay, we'll probably want to split the `backends` feature into deps which work back to oldest supported Python, and deps which don't

mdhaber:
No, I've been trying to fix the Python 3.13 requirement. It was added because `mparray` wasn't passing all `array-api-tests` with NumPy <= 2.5 (mdhaber/mparray#17):

> The `2024.12` standard specified that `meshgrid` and `broadcast_arrays` should return `list`s rather than `tuple`s. NumPy < 2.5.2 is supposed to conform to the 2024.12 version of the standard but returns `tuple`s anyway, causing the tests to fail. So I pinned NumPy >= 2.5.2, but that rules out Python 3.10/3.11 unless we want to skip the tests.

But for now, that will just mean we don't run array-api-tests in CI with Python 3.10/3.11. There's no other incompatibility, so now I'm just waiting for mparray 0.2.2 to propagate to conda-forge, then I can update this PR.

src/array_api_extra/_set.py
mdhaber: Added `is_mparray_namespace`.

tests/main/test_at.py
mdhaber: Done.
```python3
# MPArray mutation is currently a little too flexible; see mdhaber/mparray#21
```

mdhaber:
I'm happy with this if you are. I can open up issues in the mparray repo about the `as_numpy` and `__array_ufunc__` skips if you're concerned about them.

tests/main/test_testing.py
lucascolley: yeah, that seems fine to me

tests/main/test_lazy.py
lucascolley: Let's open an issue on the `mparray` side for this one, given that it isn't entirely clear which way to go yet. That would be sufficient to move forward here

lucascolley (approved):
commented about one upstream issue that would be nice, otherwise this looks great, thanks Matt!

mdhaber:
Thanks @lucascolley. Opened mdhaber/mparray#22 to remind us to consider `lazy_apply(..., as_numpy=True)`.

